PR #2, opened 2021-06-01
(feat) install docker, and allow plugins to execute

tphoney:


engine/compiler/os.go
```diff
@@ -79,3 +83,124 @@ func genScript(os string, commands []string) string {
...
+			continue
+		}
+		// all settings are passed to the plugin env
+		// variables, prefixed with PLUGIN_
+		key = "PLUGIN_" + strings.ToUpper(key)
+
+		// if the setting parameter is sources from the
+		// secret we create a secret enviornment variable.
+		if value.Secret != "" {
+			s := fmt.Sprintf(" --env %s=\"%s\"", key, value.Secret)
```
bradrydzewski, line 110: you can use `%q` to automatically escape and quote the string

PR #42, opened 2021-12-01
Refactor of VM instance pool management

marko-gacesa:


command/delegate/delegate.go
```diff
@@ -424,13 +403,10 @@ func (c *delegateCommand) handleDestroy(poolManager *vmpool.Manager) http.Handle
...
 		fmt.Printf("\n\nExecuting cleanup: %v\n", reqData)
 
-		pool := poolManager.Get(reqData.Pool)
-		instance := &vmpool.Instance{
-			ID: reqData.ID,
-			IP: "", // TODO remove this
-		}
-		destroyErr := pool.Destroy(r.Context(), instance)
-		if destroyErr != nil {
+		ctx := r.Context()
```
tphoney, line 406: nit no need for this line.

internal/vmpool/pool.go
```diff
@@ -2,13 +2,19 @@ package vmpool
 
 import (
 	"context"
+	"sync"
 )
 
+const RunnerName = "drone-runner-aws"
```
tphoney, line 8: the runner should not be a constant, it is the name of the machine that starts the runner https://github.com/drone-runners/drone-runner-aws/blob/4a3a405b58f4c1293275d2ecc8b41f926e67cba0/command/delegate/delegate.go#L60

tphoney (changes requested):

command/delegate/delegate.go
marko-gacesa: how's that? the `ctx` is used just below.

internal/vmpool/pool.go
marko-gacesa: This is not a runner name. This is actually used for the "creator" tag. It must be fixed.

command/delegate/delegate.go
tphoney: why instantiate a variable, when you can just call the method on line 408

tphoney (approved):

PR #43, opened 2021-12-02
update api setup parms

eoinmcafee00:


command/delegate/delegate.go
```diff
@@ -317,14 +307,9 @@ func (c *delegateCommand) handleSetup(poolManager *vmpool.Manager) http.HandlerF
...
 		}
-		logrus.
-			WithField("ami", pool.GetInstanceType()).
-			WithField("pool", pool.GetName()).
-			WithField("ip", instance.IP).
-			WithField("id", instance.ID).
-			WithField("response", *setupResponse).
-			Info("handleSetup: setup complete")
-		w.WriteHeader(http.StatusOK)
+		// return response
```
tphoney, line 310: can we put the logging back
```
logrus.
			WithField("ami", pool.GetInstanceType()).
			WithField("pool", pool.GetName()).
			WithField("ip", instance.IP).
			WithField("id", instance.ID).
			WithField("response", *setupResponse).
			Info("handleSetup: setup complete")
```

tphoney (changes requested):

command/delegate/delegate.go
eoinmcafee00: done

tphoney (approved):

PR #50, opened 2021-12-06
(maint) update setup struct from lite engine to internal response struct

eoinmcafee00:


command/delegate/delegate.go
```diff
@@ -30,12 +30,18 @@ import (
...
-	poolManager         *vmpool.Manager
-}
+type (
+	delegateCommand struct {
+		envfile             string
+		awsPoolfile         string
+		defaultPoolSettings vmpool.DefaultSettings
+		poolManager         *vmpool.Manager
+	}
+	setupResponse struct {
```
tphoney, line 40: can you put the setupResponse in https://github.com/drone-runners/drone-runner-aws/blob/master/command/delegate/data.go

command/delegate/delegate.go
```diff
@@ -303,12 +309,12 @@ func (c *delegateCommand) handleSetup() http.HandlerFunc {
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
-		JSON(w, api.SetupResponse{IPAddress: instance.IP, InstanceID: instance.ID}, http.StatusOK)
+		JSON(w, setupResponse{IPAddress: instance.IP, InstanceID: instance.ID}, http.StatusOK)
```
marko-gacesa, line 312:
```suggestion
		JSON(w, struct {
			InstanceID string `json:"instance_id,omitempty"`
			IPAddress  string `json:"ip_address,omitempty"`
		}{
			IPAddress: instance.IP,
			InstanceID: instance.ID
		}, http.StatusOK)
```

command/delegate/delegate.go
```diff
@@ -303,12 +309,12 @@ func (c *delegateCommand) handleSetup() http.HandlerFunc {
...
 		logrus.
 			WithField("pool", poolName).
 			WithField("ip", instance.IP).
 			WithField("id", instance.ID).
-			WithField("response", *setupResponse).
+			WithField("response", *liteEngineSetupResponse).
```
tphoney, line 317: this should stay WithField("response", *setupResponse).

tphoney (changes requested):

command/delegate/delegate.go
marko-gacesa: Because the struct does not have to exist outside of this function, this one place.

PR #55, opened 2021-12-20
daemon: run commands through LE

marko-gacesa:


engine/compiler/compiler.go
```diff
@@ -46,44 +48,60 @@ func (c *Compiler) Compile(ctx context.Context, args runtime.CompilerArgs) runti
...
+	pipelineOS := pool.GetOS()
+	pipelineRoot := pool.GetRootDir()
+
 	// move the pool from the `mapping of pools` into the spec of this pipeline.
 	spec.CloudInstance.PoolName = targetPool
-	spec.Root = pool.GetRootDir()
 
-	pipelineOS := pool.GetOS()
 	// creates a home directory in the root.
 	// note: mkdirall fails on windows so we need to create all directories in the tree.
```
shubham149, line 63: In golang, os.MkdirAll works for both windows & linux I guess.

engine/compiler/compiler.go
```diff
@@ -163,22 +183,35 @@ func (c *Compiler) Compile(ctx context.Context, args runtime.CompilerArgs) runti
...
+
+		// TODO: Remove when lespec.Step is with Files slice
+		_ = clonepath
+		_ = clonefile
+
 		cmd, args := oshelp.GetCommand(pipelineOS, clonepath)
 		spec.Steps = append(spec.Steps, &engine.Step{
-			Name:      "clone",
+			Step: lespec.Step{
+				ID:   random(),
```
shubham149, line 194: Also, set a Key field in all the steps. It can be same as ID.

marko-gacesa: What is the exact name of the field? I can't find it...

shubham149: https://github.com/harness/lite-engine/blob/main/api/api.go#L47

shubham149: LogKey

tphoney (approved):

PR #70, opened 2022-02-03
Use AWS IAM role to create VMs

shubham149:


internal/vmpool/cloudaws/config_file.go
```diff
@@ -113,7 +113,7 @@ func ProcessPoolFile(rawFile string, defaultPoolSettings *vmpool.DefaultSettings
 
 		// we need Access, error if its still empty
 		if poolDef.Account.AccessKeyID == "" {
-			return nil, errors.New("missing AWS access key. Add to .env file or pool file")
+			logrus.Infof("AWS access key is not provided (falling back to ec2 instance profile)")
```
tphoney, line 116: so the IAM role authentication relies on not passing credentials. 
if that is how it works, we should make sure that we check that iamProfileArn is populated. 

If both are not set we should return with an error

tphoney (changes requested):

internal/vmpool/cloudaws/config_file.go
shubham149: When we attach an IAM role to the VM on which runner is executing on, it does not need anything to be passed. Even iamProfileArn doesn't have to be passed in that case.

shubham149: If we want assume role and user role arn support as well, we can do something like:
`func (c *Creds) GetClient() (*ec2.EC2, error) {
	// create the client
	conf := &aws.Config{
		Region: aws.String(c.Region),
	}

	if c.AccessKey != "" && c.SecretKey != "" {
		conf.Credentials = credentials.NewStaticCredentials(c.AccessKey, c.SecretKey, "")
	} else if c.AssumeRole != "" {
		conf.Credentials = getAssumeRoleCred(c.AssumeRole, c.AssumeRoleSessionName)
	} else {
		log.Warn("AWS Key and/or Secret not provided (falling back to ec2 instance profile)")
	}

	sess, err := session.NewSession(conf)
	if err != nil {
		log.WithError(err).Errorln("could not instantiate session")
		return nil, err
	}

	var client *ec2.EC2
	// If user role ARN is set then assume role here
	if len(c.UserRoleArn) > 0 {
		confRoleArn := aws.Config{
			Region:      aws.String(c.Region),
			Credentials: stscreds.NewCredentials(sess, c.UserRoleArn),
		}

		client = ec2.New(sess, &confRoleArn)
	} else {
		client = ec2.New(sess)
	}
	return client, nil
}
`

This is used by s3 upload plugin as well.

tphoney: sounds good.

shubham149:
This change adds support for IAM role to manage pool instances on aws drone runner. To use the IAM role, aws runner needs to run on EC2 instance with [IAM](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html) role having CRUD permissions on EC2. This will allow the runner to use the instance’s IAM role to get temporary security credentials to make calls to AWS for managing pool & removes requirement of specifying DRONE_SETTINGS_AWS_ACCESS_KEY_ID and DRONE_SETTINGS_AWS_ACCESS_KEY_SECRET

PR #74, opened 2022-02-09
(DRON-220) add availability zones for aws

tphoney:


eoinmcafee00 (approved):

internal/vmpool/cloudaws/pool.go
```diff
@@ -35,9 +35,10 @@ type awsPool struct {
 	privateKey    string
 	iamProfileArn string
 
-	region  string
-	os      string
-	rootDir string
+	region           string
+	AvailabilityZone string
```
marko-gacesa, line 39: maybe change this to a private field and create a getter method for it to match the implementation of the other fields.

tphoney: made the fix

marko-gacesa (approved):

PR #78, opened 2022-02-23
(feat) gcp support

eoinmcafee00:


command/delegate/delegate.go
```diff
@@ -123,7 +123,14 @@ func (c *delegateCommand) run(*kingpin.ParseContext) error {
 		KeyFile:             config.DefaultPoolSettings.KeyFile,
 	}
 	// process the pool file
-	pools, poolFileErr := cloudaws.ProcessPoolFile(c.awsPoolfile, &c.defaultPoolSettings)
+	//pools, poolFileErr := cloudaws.ProcessPoolFile(c.awsPoolfile, &c.defaultPoolSettings)
+	//if poolFileErr != nil {
```
eoinmcafee00, line 127: ignore for now, will be fixed when we update to support one pool file

internal/vmpool/cloudaws/pool.go
```diff
@@ -74,10 +74,6 @@ func (p *awsPool) GetOS() string {
 	return p.os
 }
 
-func (p *awsPool) GetUser() string {
-	return p.user
```
eoinmcafee00, line 78: Didn't see any evidence this was being used.

command/daemon/daemon.go
```diff
@@ -125,7 +127,13 @@ func (c *daemonCommand) run(*kingpin.ParseContext) error {
 		CertFile:            config.DefaultPoolSettings.CertFile,
 		KeyFile:             config.DefaultPoolSettings.KeyFile,
 	}
-	pools, err := cloudaws.ProcessPoolFile(c.poolFile, &defaultPoolSettings)
+	//pools, err := cloudaws.ProcessPoolFile(c.poolFile, &defaultPoolSettings)
```
marko-gacesa, line 130: There is no need to comment these lines out. The `poolManager` supports multiple pools. Use something like:
```
poolsAWS, err := cloudaws.ProcessPoolFile(c.poolFile, &defaultPoolSettings)
err = poolManager.Add(poolsAWS...)

poolsGCP, err := google.ProcessPoolFile(c.googlePoolFile, &defaultPoolSettings)
err = poolManager.Add(poolsGCP...)
```

We'll refactor it later when we switch to using a single pool file solution.

command/delegate/delegate.go
marko-gacesa: yes, this is the correct implementation for now.. see my comment in the daemon.go

internal/vmpool/google/config.go
```diff
@@ -0,0 +1,256 @@
+package google
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/drone-runners/drone-runner-aws/internal/cloudinit"
```
marko-gacesa, line 9: order the imports correctly...

command/delegate/delegate.go
eoinmcafee00: I've uncommented this for now - will pick up work after merge

internal/vmpool/google/config.go
```diff
@@ -0,0 +1,258 @@
...
+
+	"github.com/drone-runners/drone-runner-aws/oshelp"
+
+	"github.com/buildkite/yaml"
+	"github.com/drone/runner-go/logger"
+
+	"github.com/drone-runners/drone-runner-aws/internal/vmpool"
```
marko-gacesa, line 16: move this line up to the correct section.

internal/vmpool/google/pool.go
```diff
@@ -0,0 +1,507 @@
...
+	if list.Items == nil {
+		return
+	}
+
+	for _, vm := range list.Items {
+		logr.
+			WithField("machine", vm.Name).
+			WithField("id", vm.Id).
+			WithField("status", vm.Status).
+			Traceln("-------- list machines -----------")
```
marko-gacesa, line 117: Maybe remove logging of every instance, every time the List method is called...?

internal/vmpool/google/config.go
eoinmcafee00: done

marko-gacesa (approved):

PR #79, opened 2022-02-23
Added support for private repos and custom entrypoint

marko-gacesa:


engine/compiler/compiler.go
```diff
@@ -54,6 +63,7 @@ func (c *Compiler) Compile(ctx context.Context, args runtime.CompilerArgs) runti
 	targetPool := pipeline.Pool.Use
 	pool := c.PoolManager.Get(targetPool)
 	if pool == nil {
+		// TODO return an error to the caller if the requested pool does not exist.
```
eoinmcafee00, line 66: Is this still to do?

engine/compiler/compiler.go
```diff
@@ -273,24 +291,58 @@ func (c *Compiler) Compile(ctx context.Context, args runtime.CompilerArgs) runti
...
 		})
 	}
 
+	// get registry credentials from registry plugins
+	creds, err := c.Registry.List(ctx, &registry.Request{
+		Repo:  args.Repo,
+		Build: args.Build,
+	})
+	if err != nil { //nolint:staticcheck
+		// TODO (bradrydzewski) return an error to the caller if the provider returns an error.
```
eoinmcafee00, line 319: Doesn't seem much point to this statement?

eoinmcafee00 (changes requested):
Just a couple of comments

engine/compiler/compiler.go
marko-gacesa: See the comment above... This particular code section is copied from the docker runner along with the "to do" comment.

marko-gacesa:
The "to do"s must stay for now.

eoinmcafee00 (approved):

PR #88, opened 2022-03-15
Add support for DRONE_RUNNER_VOLUMES in the Drone AWS runner.

vistaarjuneja:
This support already exists in the docker and k8s runners.
Ref: https://docs.drone.io/runner/nomad/configuration/reference/drone-runner-volumes/
Sample PR for the k8s runner: https://github.com/drone-runners/drone-runner-kube/pull/24

In the above implementation for k8s, map is used to represent volumes but there can be duplicate sources (eg source 'a' mapping to destinations 'b' and 'c'). Hence changed that to a list here.
Will fix this in the docker and kubernetes runners as well.

### DIFF
--- a/command/compile.go
+++ b/command/compile.go
@@ -37,6 +37,7 @@ type compileCommand struct {
 	Environ map[string]string
 	Secrets map[string]string
 	Config  string
+	Volumes []string
 }
 
 func (c *compileCommand) run(*kingpin.ParseContext) error {
@@ -116,6 +117,7 @@ func (c *compileCommand) run(*kingpin.ParseContext) error {
 		Environ:     provider.Static(c.Environ),
 		NetworkOpts: nil,
 		Secret:      secret.StaticVars(c.Secrets),
+		Volumes:     c.Volumes,
 		PoolManager: poolManager,
 		Registry:    registry.File(c.Config),
 	}
@@ -155,6 +157,9 @@ func registerCompile(app *kingpin.Application) {
 
 	cmd.Flag("secrets", "secret parameters").
 		StringMapVar(&c.Secrets)
+
+	cmd.Flag("volumes", "drone runner volumes").
+		StringsVar(&c.Volumes)
 
 	cmd.Flag("environ", "environment variables").
 		StringMapVar(&c.Environ)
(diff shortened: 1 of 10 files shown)


command/delegate/delegate.go
```diff
@@ -428,6 +469,32 @@ func (c *delegateCommand) handleStep(w http.ResponseWriter, r *http.Request) {
...
 	ctx := r.Context()
 
+	// load the configuration from the environment
+	config, err := fromEnviron()
+	if err != nil {
+		return
+	}
+
+	// add global volumes as mounts only if image is specified
+	if reqData.Image != "" {
```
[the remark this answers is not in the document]
shubham149: how will they use global certs?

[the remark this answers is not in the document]
vistaarjuneja: those need to be present in the ami or via the cloud init

shubham149:
Volumes is a map in drone docker. We cannot convert it to list now since that will break backward compatibility.

command/compile.go
```diff
@@ -156,6 +158,9 @@ func registerCompile(app *kingpin.Application) {
 	cmd.Flag("secrets", "secret parameters").
 		StringMapVar(&c.Secrets)
 
+	cmd.Flag("volumes", "drone runner volumes").
```
shubham149, line 161: Add some more doc on what's this volume and how it is used?

command/delegate/delegate.go
```diff
@@ -318,6 +323,35 @@ func (c *delegateCommand) handleSetup(w http.ResponseWriter, r *http.Request) {
...
+	// load the configuration from the environment
+	env, err := fromEnviron()
+	if err != nil {
+		log.Errorf("could not load config from environment: %s", err)
+		return
+	}
+
+	// append global volumes to the setup request.
+	for _, pair := range env.Runner.Volumes {
+		plen := 2
```
shubham149, line 335: L335-L343 is used in 3 places. Let's create a method for it that probably returns src & ro & use it instead of duplicating the code.

command/compile.go
vistaarjuneja: added a comment for reference

command/delegate/delegate.go
vistaarjuneja: done

shubham149 (approved):

PR #89, opened 2022-03-18
[POC]: Add support for vm hibernate to reduce costs

shubham149:
Currently, pool size number of aws runner vms are always running which increases the build cost. This POC pr takes the approach to hibernate the vm after its bootup is complete. Hibernate is similar to closing the lid of laptop which is restored to earlier state within seconds after starting the instance back. Advantage of hibernated vm is ec2 does not charge for compute of hibernated (stopped) vms. Only storage is charged. This will reduce the cost to use aws runner by upto 90%.

### DIFF
--- a/internal/vmpool/manager.go
+++ b/internal/vmpool/manager.go
@@ -231,6 +231,12 @@ func (m *Manager) buildPool(ctx context.Context, pool *poolEntry) error {
 				WithField("pool", pool.GetName()).
 				WithField("id", instance.ID).
 				Infoln("build pool: created new instance")
+			go func() {
+				herr := m.Hibernate(ctx, pool.GetName(), instance.ID)
+				if herr != nil {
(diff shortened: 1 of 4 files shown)


internal/vmpool/manager.go
```diff
@@ -231,6 +232,12 @@ func (m *Manager) buildPool(ctx context.Context, pool *poolEntry) error {
 				WithField("pool", pool.GetName()).
 				WithField("id", instance.ID).
 				Infoln("build pool: created new instance")
+			go func() {
+				herr := m.Hibernate(ctx, pool.GetName(), instance.ID)
```
eoinmcafee00, line 236: This functionality should be configurable? It's also not going to be supported across all providers, so a check that the provider is AWS otherwise it will throw a not supported error in the future.

shubham149: This is just a POC. It is not supposed to be merged in prod.

PR #91, opened 2022-04-07
Add support for VM hibernate

shubham149:


internal/drivers/amazon/pool.go
```diff
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
```
eoinmcafee00, line 10: import ordering

shubham149: Done

eoinmcafee00 (approved):

PR #96, opened 2022-04-29
Add stage owner support in pool ownership api

shubham149:
Currently only one runner can be the owner of a specific pool in harness CIE. For OSX builds, this is a big limitation since only 2 parallel builds can happen for a pipeline.
To fix this limitation, this change is required. Without this change, cleanup is not guaranteed to happen. Pipeline can be aborted at any time. In case it is aborted just a second after starting up, setup & cleanup call can happen in parallel. We use capability check in CIE to determine which delegate needs to execute a VM setup task. If more than 2 runners are responsible for same pool, stage ownership is added to pool owner api so that cleanup call gets assigned to correct delegate.

### DIFF
--- a/store/database/migrate/postgres/0002_create_table_instances_stageowner.up.sql
+++ b/store/database/migrate/postgres/0002_create_table_instances_stageowner.up.sql
@@ -1,0 +1,28 @@
+CREATE TABLE IF NOT EXISTS instances (
+     instance_id        VARCHAR(250) PRIMARY KEY
+    ,instance_name      VARCHAR(250)
+    ,instance_address   VARCHAR(250)
+    ,instance_provider  VARCHAR(50)
+    ,instance_state     VARCHAR(50)
+    ,instance_pool      VARCHAR(250)
+    ,instance_image     VARCHAR(250)
+    ,instance_region    VARCHAR(50)
+    ,instance_zone      VARCHAR(50)
+    ,instance_size      VARCHAR(50)
+    ,instance_platform  VARCHAR(50)
+    ,instance_arch      VARCHAR(50)
+    ,instance_stage     INTEGER
+    ,instance_ca_key    TEXT
+    ,instance_ca_cert   TEXT
+    ,instance_tls_key   TEXT
+    ,instance_tls_cert  TEXT
+    ,instance_updated   INTEGER
+    ,instance_started   INTEGER
+    ,is_hibernated      BOOLEAN
+,UNIQUE(instance_name)
+);
+
+CREATE TABLE IF NOT EXISTS stage_owner (
+     stage_id          VARCHAR(250) PRIMARY KEY
+    ,pool_name         VARCHAR(250)
+);
--- a/store/database/migrate/postgres/0002_create_table_instances_stageowner.up.sql
+++ b/store/database/migrate/postgres/0002_create_table_instances_stageowner.up.sql
@@ -1,27 +1,3 @@
-CREATE TABLE IF NOT EXISTS instances (
-     instance_id        VARCHAR(250) PRIMARY KEY
-    ,instance_name      VARCHAR(250)
-    ,instance_address   VARCHAR(250)
-    ,instance_provider  VARCHAR(50)
-    ,instance_state     VARCHAR(50)
-    ,instance_pool      VARCHAR(250)
-    ,instance_image     VARCHAR(250)
-    ,instance_region    VARCHAR(50)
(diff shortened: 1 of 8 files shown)


tphoney (changes requested):
am not comfortable merging this for now, @Eoin Mcafee would be a better person to look at this db change. i havent touched that area at all yet. 
one thing that is missing is migration scripts for the db, which were missing on your last pr too.
in plain english, "this adds an owner on start of setup, then checks that when registering with the delegate ?"

store/database/migrate/postgres/0002_create_table_instances_stageowner.up.sql
```diff
@@ -0,0 +1,28 @@
+CREATE TABLE IF NOT EXISTS instances (
```
eoinmcafee00, line 1: These files should only contain new db changes. Like creating a table or amending a table to add a new column. Instance creation is already done in the previous change scipt

eoinmcafee00:
Hey @shubham149 - I'd like @bradrydzewski to take a look at this PR before approving. I'm not entirely comfortable with having a new table just to support delegate. He may be able see a better approach.

store/database/migrate/postgres/0002_create_table_instances_stageowner.up.sql
shubham149: Done

eoinmcafee00: I think you've misunderstood how the migration scripts work. The 1st script 001 has already been run. therefore you don't need to include the instance table as it was already created. If it's the first time you're running the app, then it will run the first and then the 2nd. 

If you need to make changes to the instance table it will be an AMEND. You wouldn't change the 1st script 001 but add a new one to amend the existing table.

shubham149: Done

eoinmcafee00:
I've been thinking about this change & I think it would be better not to have another new table just for CIE. I'd like to propose we keep just the existing table & add the ability to create an instance record at the point you need to with poolName & StageId & marry them up when we fill in the rest of the details? What are your thoughts @bradrydzewski ?

bradrydzewski:
@eoinmcafee00 sounds good ... I think we do something similar with the autoscaler, and certainly would be great to limit to a single table if possible.  Let's explore your suggestion to see if it solves for Shubham's use case.

shubham149:
@eoinmcafee00 This is a must for 2 runners managing a pool with same name. Context of setup call is not passed to destroy call. 

There are couple of issues in case we use a single table. Major one being:
1. Instances are provisioned by the background pool thread. So, db entity is already present for it. We require an entity to store the mapping of stage id and pool id before an instance provision call is done. So, if setup call creates a dummy entity before instance provision call, how will we merge the two entities. It will be simpler to use 2 tables in this use case.

eoinmcafee00:
"So, if setup call creates a dummy entity before instance provision call, how will we merge the two entities." - we would be storing pool & stage which is already present before the other details. This is the same details you are storing in this PR just in a different table?

shubham149:
Stage ID is provided at the setup call and not as part of pool file. Hence, background pool thread cannot set the stage id for the created instances.

eoinmcafee00:
Can you not figure this out using https://github.com/drone-runners/drone-runner-aws/blob/master/command/config/config.go#L164 ? 
Can you provide screenshots of this change working?

tphoney (changes requested):

types/types.go
```diff
@@ -79,3 +79,8 @@ type QueryParams struct {
 	Status InstanceState
 	Stage  string
 }
+
+type StageOwner struct {
+	ID   string `db:"stage_id" json:"stage_id"`
```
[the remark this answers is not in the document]
shubham149: addressed the comment

types/types.go
```diff
@@ -79,3 +79,8 @@ type QueryParams struct {
 	Status InstanceState
 	Stage  string
 }
+
+type StageOwner struct {
+	StageID string `db:"stage_id" json:"stage_id"`
+	Pool    string `db:"pool_name" json:"pool_name"`
```
tphoney, line 85: sorry and pool/poolname.

tphoney (changes requested):

types/types.go
shubham149: Addressed the comment

PR #98, opened 2022-05-04
Add dynamic wait for vm hibernate

shubham149:


internal/drivers/manager.go
```diff
@@ -2,17 +2,19 @@ package drivers
...
 	"sort"
 	"sync"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/drone-runners/drone-runner-aws/internal/certs"
 	"github.com/drone-runners/drone-runner-aws/store"
 	"github.com/drone-runners/drone-runner-aws/types"
 	"github.com/drone/runner-go/logger"
+	lehttp "github.com/harness/lite-engine/cli/client"
```
eoinmcafee00, line 16: import ordering

shubham149: I guess ordering is already correct. No changes are shown on go fmt

internal/drivers/manager.go
```diff
@@ -2,17 +2,19 @@ package drivers
...
 import (
 	"context"
-	"errors"
 	"fmt"
 	"runtime/debug"
 	"sort"
 	"sync"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
```
eoinmcafee00, line 11: this should be below drone imports in a separate section. See how other files do it in the repo

eoinmcafee00: because it's 3rd party

eoinmcafee00: ignore you've already done it

eoinmcafee00 (approved):

PR #99, opened 2022-05-04
(feat) - new provider for building on osx - Anka

eoinmcafee00:
https://docs.veertu.com/anka/intel/

Delegate testing:

Screenshot 2022-05-12 at 11 25 41
Screenshot 2022-05-12 at 11 17 51
Screenshot 2022-05-12 at 11 15 49

Harness tests:
https://harness.drone.io/eoinmcafee00/HelloWorldCore


command/config/config.go
```diff
@@ -97,8 +97,18 @@ type (
...
 		RootDirectory string `json:"root_directory,omitempty" yaml:"root_directory"`
 	}
+	// Anka specifies the configuration for an Anka instance.
+	Anka struct {
+		Account struct {
+			Username string `json:"username,omitempty"  yaml:"username"`
+			Password string `json:"password,omitempty"  yaml:"password"`
+		}
+		VMID          string `json:"vm_id,omitempty" yaml:"vm_id"`
+		RootDirectory string `json:"root_directory,omitempty" yaml:"root_directory"`
```
tphoney, line 107: what is this the root dir for, the binary ?

internal/drivers/anka/pool.go
```diff
@@ -0,0 +1,237 @@
...
+
+	logr.Info("Running script in VM")
+
+	cmdRunScript := commandRunScript(createdVM.Body.UUID, f.Name())
+	_, err = cmdRunScript.CombinedOutput()
+	if err != nil {
+		return nil, err
+	}
+
+	startTime := time.Now()
```
tphoney, line 127: should move to the top, of the function

internal/drivers/anka/pool.go
```diff
@@ -0,0 +1,237 @@
...
+	return nil
+}
+
+func (p *provider) CanHibernate() bool {
+	return false
+}
+
+func (p *provider) Create(ctx context.Context, opts *types.InstanceCreateOpts) (instance *types.Instance, err error) {
+	uData := lehelper.GenerateUserdata(p.userData, opts)
+	machineName := fmt.Sprintf(opts.RunnerName+"-"+"-%d", time.Now().Unix())
```
tphoney, line 51: use startTime in here

internal/drivers/anka/pool.go
```diff
@@ -0,0 +1,237 @@
...
+		return nil, err
+	}
+
+	cmdStartVM := commandAnka(machineName, "start")
+	_, err = cmdStartVM.CombinedOutput()
+	if err != nil {
+		return nil, err
+	}
+	var ip string
+	for i := 1; i <= 60; i++ {
```
tphoney, line 71: 60 seems like a magic number

internal/drivers/anka/pool.go
```diff
@@ -0,0 +1,237 @@
...
+func (p *provider) RootDir() string {
+	return p.rootDir
+}
+
+func (p *provider) ProviderName() string {
+	return string(types.ProviderAnka)
+}
+
+func (p *provider) Ping(_ context.Context) error {
+	return nil
```
tphoney, line 42: ping is needed

internal/drivers/anka/pool.go
```diff
@@ -0,0 +1,237 @@
...
+
+	"github.com/drone-runners/drone-runner-aws/internal/lehelper"
+	"github.com/drone-runners/drone-runner-aws/types"
+	"github.com/drone/runner-go/logger"
+
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	ankabin = "/usr/local/bin/anka"
```
tphoney, line 22: if this never changes it should not be a var

internal/drivers/anka/pool.go
```diff
@@ -0,0 +1,237 @@
...
+
+func (p *provider) Ping(_ context.Context) error {
+	return nil
+}
+
+func (p *provider) CanHibernate() bool {
+	return false
+}
+
+func (p *provider) Create(ctx context.Context, opts *types.InstanceCreateOpts) (instance *types.Instance, err error) {
```
tphoney, line 49: there is little or no context given in error messages when there is a failure in this method, we just return err. 
it would be helpful to explain what we were trying to do when we failed.

tphoney (changes requested):

internal/drivers/anka/pool.go
eoinmcafee00: ping what? The app is installed or its not.

eoinmcafee00: there's nothing to "ping"

internal/drivers/anka/pool.go
eoinmcafee00: There's no point in making this configurable - usually only takes about 2-3 secs to get an IP then exits

internal/drivers/anka/pool.go
eoinmcafee00: done

internal/drivers/anka/pool.go
eoinmcafee00: done

internal/drivers/anka/pool.go
eoinmcafee00: done

internal/drivers/anka/pool.go
eoinmcafee00: added additional logging

internal/drivers/anka/pool.go
tphoney: as a bare minimum i would expect we check that the binary for anka is available.

internal/drivers/anka/pool.go
tphoney: can you put in a comment to explaining what the number is for

tphoney (changes requested):
Still a couple of things. 
And you need to acceptance tests showing that the aws runner still works for drone, using the daemon command. 
also have the runner work in delegate mode and you can hit the end points. 
documented here https://github.com/drone-runners/drone-runner-aws/blob/master/.github/pull_request_template.md
this is until we get automated testing

internal/drivers/anka/pool.go
eoinmcafee00: done

tphoney (approved):

PR #112, opened 2022-05-23
(feat) provides ability to start a mac instance without docker to support m1

eoinmcafee00:


tphoney (changes requested):
Looks good, however it would be good to show this working in a screen shot. 
And there is no acceptance tests for delegate.

PR #119, opened 2022-05-27
(feat) turn off auto docker volume mount if ARM64 as nested virtualisation isn't support on m1

eoinmcafee00:
Only change it for drone because right now delegate does not support arm64


engine/engine.go
```diff
@@ -128,6 +130,10 @@ func (eng *Engine) Setup(ctx context.Context, specv runtime.Spec) error {
 		Files:     spec.Files,
 	}
 
+	if instance.Arch == ARM64 {
```
tphoney, line 133: can you put `const ArchARM64 = "arm64"` into oshelp similar to https://github.com/drone-runners/drone-runner-aws/pull/117/files#diff-2eb7297b270d1432e87ed95c084b3db7095e82eb638e3ae8e52e30af4f658774

tphoney (changes requested):

engine/engine.go
eoinmcafee00: done

tphoney (changes requested):
this needs to be paired with os type or it will disable docker for ubuntu arm64

engine/engine.go
```diff
@@ -128,6 +130,10 @@ func (eng *Engine) Setup(ctx context.Context, specv runtime.Spec) error {
 		Files:     spec.Files,
 	}
 
+	if instance.Arch == oshelp.ArchArm64 {
```
shubham149, line 133: Add a check that os needs to be darwin

engine/engine.go
```diff
@@ -128,6 +130,10 @@ func (eng *Engine) Setup(ctx context.Context, specv runtime.Spec) error {
 		Files:     spec.Files,
 	}
 
+	if instance.Platform == oshelp.OSMac && instance.Arch == oshelp.ArchArm64 {
```
tphoney, line 133: can you put a comment in here, to explain what is happening. for future peeps

tphoney: you are explaining the code in the line below, you do not explain why.

```
// Currently the OSX m1 architecture does not enable nested virtualisation, so we disable docker.
```

PR #121, opened 2022-06-01
(DRON-324) simplify setup for google

tphoney:
# Commit Checklist

## The Basics

### DIFF
--- a/internal/drivers/google/option.go
+++ b/internal/drivers/google/option.go
@@ -10,7 +10,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-type Option func(*provider)
+type Option func(*config)
 
 func SetPlatformDefaults(platform *types.Platform) (*types.Platform, error) {
 	if platform.Arch == "" {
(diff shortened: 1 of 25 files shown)


internal/drivers/google/option.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-type Option func(*provider)
+type Option func(*googleConfig)
```
eoinmcafee00, line 13: there's no need for this to be googleConfig - you're already in the package "google" so "config" is sufficient

eoinmcafee00: same for the rest of them.

internal/drivers/google/pool.go
```diff
@@ -65,7 +64,7 @@ func (p *provider) Create(ctx context.Context, opts *types.InstanceCreateOpts) (
 	zone := p.Zone()
 
 	logr := logger.FromContext(ctx).
-		WithField("cloud", types.ProviderGoogle).
+		WithField("cloud", types.DriverGoogle).
```
eoinmcafee00, line 67: There's no need to append "Driver" google should be sufficient

eoinmcafee00 (approved):

PR #139, opened 2022-07-14
(feat) azure support for linux

eoinmcafee00:


internal/drivers/azure/driver.go
```diff
@@ -0,0 +1,344 @@
...
+	publicIPName := fmt.Sprintf("%s-publicip", name)
+	networkInterfaceName := fmt.Sprintf("%s-networkinterface", name)
+	diskName := fmt.Sprintf("%s-disk", name)
+
+	var tags = map[string]*string{}
+	// add user defined tags
+	var t string
+	for k, v := range c.tags {
+		t = v
+		tags[k] = &t
```
vistaarjuneja, line 120: wouldn't tags[k] always point to the same address here? &t would always be the same, right?

internal/drivers/azure/driver.go
```diff
@@ -0,0 +1,344 @@
...
+	logr := logger.FromContext(ctx).
+		WithField("cloud", types.Azure).
+		WithField("name", name).
+		WithField("image", c.InstanceType()).
+		WithField("pool", opts.PoolName).
+		WithField("zone", "").
+		WithField("image", c.offer).
+		WithField("size", c.size)
+
+	logr.Debugln("Starting Azure Setup")
```
vistaarjuneja, line 132: Maybe better to have this in info mode rather than debug (if it's not printed anywhere else?)

internal/drivers/azure/driver.go
```diff
@@ -0,0 +1,344 @@
...
+		}
+	})
+
+	if c.resourceGroupName == "" {
+		c.resourceGroupName = defaultResourceGroup
+	}
+
+	_, err = c.createVirtualNetwork(ctx, c.cred, vnetName)
+	if err != nil {
+		logr.Errorln(err)
```
vistaarjuneja, line 148: Would it be better to do something like `logr.WithError(err).Error("could not create virtual network")` if the error is not descriptive enough?  (and likewise for the other calls below)

internal/drivers/azure/driver.go
```diff
@@ -0,0 +1,344 @@
...
+	if err != nil {
+		return nil, err
+	}
+	vm, err := poller.PollUntilDone(ctx, nil)
+	if err != nil {
+		return nil, err
+	}
+	logr.Debugln("instance insert operation completed")
+
+	logr.
```
vistaarjuneja, line 233: Nit: format

internal/drivers/azure/driver.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+func (c *config) Destroy(ctx context.Context, instanceIDs ...string) (err error) {
+	logr := logger.FromContext(ctx)
+	if c.resourceGroupName == "" {
+		c.resourceGroupName = defaultResourceGroup
+	}
+	logr.Debugln("Azure destroy operation started")
+	if len(instanceIDs) == 0 {
+		return
```
vistaarjuneja, line 254: not returning anything here

internal/drivers/azure/driver.go
```diff
@@ -0,0 +1,344 @@
...
+
+		poller, err := c.service.BeginDelete(ctx, c.resourceGroupName, instanceID, nil)
+		if err != nil {
+			return err
+		}
+		_, err = poller.PollUntilDone(ctx, nil)
+		if err != nil {
+			return err
+		}
+		logr.Debugf("Azure instance destroyed %s", instanceID)
```
vistaarjuneja, line 274: just leaving it here: maybe better to convert debug logs to info logs if you feel they are insightful

internal/drivers/azure/nic.go
```diff
@@ -0,0 +1,241 @@
...
+func (c *config) createSubnets(ctx context.Context, cred azcore.TokenCredential, subnetName, vnetName string) (*armnetwork.Subnet, error) {
+	logr := logger.FromContext(ctx)
+
+	subnetClient, err := armnetwork.NewSubnetsClient(c.subscriptionID, cred, nil)
+	if err != nil {
+		return nil, err
+	}
+	parameters := armnetwork.Subnet{
+		Properties: &armnetwork.SubnetPropertiesFormat{
+			AddressPrefix: to.Ptr("10.1.10.0/24"),
```
vistaarjuneja, line 104: consider moving these kinds of strings to top of the file and defining in variables

internal/drivers/azure/driver.go
eoinmcafee00: good spot - fixed

internal/drivers/azure/driver.go
eoinmcafee00: done

internal/drivers/azure/driver.go
eoinmcafee00: updated.

internal/drivers/azure/driver.go
eoinmcafee00: returning value

internal/drivers/azure/driver.go
eoinmcafee00: yeah changed.

internal/drivers/azure/nic.go
eoinmcafee00: done.

internal/drivers/azure/driver.go
```diff
@@ -0,0 +1,346 @@
...
+		WithField("image", c.InstanceType()).
+		WithField("pool", opts.PoolName).
+		WithField("zone", "").
+		WithField("image", c.offer).
+		WithField("size", c.size)
+
+	logr.Info("Starting Azure Setup")
+
+	c.init.Do(func() {
+		_, err = c.createResourceGroup(ctx, c.cred)
```
vistaarjuneja, line 136: is the creation of resource group the only call that should be in the `init.Do`?

eoinmcafee00: updated

vistaarjuneja (approved):

PR #140, opened 2022-07-15
added clean up on provision failure

eoinmcafee00:
If provisioning fails of an instance - it will clean up the stage owner table

### DIFF
--- a/command/harness/destroy.go
+++ b/command/harness/destroy.go
@@ -42,9 +42,12 @@ func HandleDestroy(ctx context.Context, r *VMCleanupRequest, s store.StageOwnerS
 		return fmt.Errorf("annot destroy the instance: %w", err)
 	}
 	logr.Traceln("destroyed instance")
-
-	if err := s.Delete(ctx, r.StageRuntimeID); err != nil {
-		logrus.WithError(err).Errorln("failed to delete stage owner entity")
+	exists, _ := s.Find(ctx, r.StageRuntimeID, r.PoolID)
+	if exists != nil {
+		if deleteErr := s.Delete(ctx, r.StageRuntimeID); err != nil {
+			logrus.WithError(deleteErr).Errorln("failed to delete stage owner entity")
+		}
 	}
+
 	return nil
 }
--- a/command/harness/destroy.go
+++ b/command/harness/destroy.go
@@ -39,13 +39,13 @@ func HandleDestroy(ctx context.Context, r *VMCleanupRequest, s store.StageOwnerS
 		WithField("pool", r.PoolID)
 
 	if err = poolManager.Destroy(ctx, r.PoolID, inst.ID); err != nil {
-		return fmt.Errorf("annot destroy the instance: %w", err)
+		return fmt.Errorf("cannot destroy the instance: %w", err)
 	}
 	logr.Traceln("destroyed instance")
 	exists, _ := s.Find(ctx, r.StageRuntimeID, r.PoolID)
 	if exists != nil {
-		if deleteErr := s.Delete(ctx, r.StageRuntimeID); err != nil {
-			logrus.WithError(deleteErr).Errorln("failed to delete stage owner entity")
+		if err = s.Delete(ctx, r.StageRuntimeID); err != nil {
+			logr.WithError(err).Errorln("failed to delete stage owner entity")
 		}
 	}
 
--- a/command/harness/destroy.go
+++ b/command/harness/destroy.go
@@ -38,7 +38,7 @@ func HandleDestroy(ctx context.Context, r *VMCleanupRequest, s store.StageOwnerS
 		WithField("stage_runtime_id", r.StageRuntimeID).
 		WithField("pool", r.PoolID)
 
-	if err := poolManager.Destroy(ctx, r.PoolID, inst.ID); err != nil {
+	if err = poolManager.Destroy(ctx, r.PoolID, inst.ID); err != nil {
 		return fmt.Errorf("annot destroy the instance: %w", err)
 	}
 	logr.Traceln("destroyed instance")
--- a/command/harness/setup.go
+++ b/command/harness/setup.go
@@ -93,21 +93,15 @@ func HandleSetup(ctx context.Context, r *SetupVMRequest, s store.StageOwnerStore
 	if !poolManager.Exists(pool) {
 		return nil, fmt.Errorf("pool not defined")
 	}
-
-	cleanUpFn := func() {
-		errCleanUpStageOwner := s.Delete(context.Background(), stageRuntimeID)
-		if errCleanUpStageOwner != nil {
-			logr.WithError(errCleanUpStageOwner).Errorln("failed to delete failed stage owner")
+	_, err := s.Find(ctx, stageRuntimeID, pool)
+	if err != nil {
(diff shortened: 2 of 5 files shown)


command/harness/setup.go
```diff
@@ -97,30 +94,43 @@ func HandleSetup(ctx context.Context, r *SetupVMRequest, s store.StageOwnerStore
...
+	}
+
+	if err := s.Create(ctx, &types.StageOwner{StageID: stageRuntimeID, PoolName: pool}); err != nil {
+		return nil, fmt.Errorf("could not create stage owner entity: %w", err)
+	}
+
 	instance, err := poolManager.Provision(ctx, pool, env.Runner.Name, env.Settings.LiteEnginePath)
 	if err != nil {
+		go cleanUpFn()
 		return nil, fmt.Errorf("failed provisioning")
```
vistaarjuneja, line 111: Let's add the error in the return here. I saw "failed provisioning" in the logs but it doesn't print the error

command/harness/setup.go
```diff
@@ -97,30 +94,43 @@ func HandleSetup(ctx context.Context, r *SetupVMRequest, s store.StageOwnerStore
...
-	}
-
 	logr = logr.
 		WithField("ip", instance.Address).
 		WithField("id", instance.ID)
 
 	// cleanUpFn is a function to terminate the instance if an error occurs later in the handleSetup function
-	cleanUpFn := func() {
+	cleanUpFn = func() {
 		errCleanUp := poolManager.Destroy(context.Background(), pool, instance.ID)
```
vistaarjuneja, line 120: should we add the cleanup for the db mapping here as well? eg what happens if call fails later - the mapping would still remain right?

In the logs, I saw one of the executions failed with the error on L152. If that happens, we still want to remove the mapping so that it successfully retries right?

eoinmcafee00: It does happen on the harness destroy. The reason it’s not been added at the manager level is because the table isn’t used by drone. It’s impl is unique to delegate. I do think maybe we should add it at manager level but it’s a bigger refactor with higher chance of regression

command/harness/setup.go
```diff
@@ -97,30 +94,43 @@ func HandleSetup(ctx context.Context, r *SetupVMRequest, s store.StageOwnerStore
...
 	}
 
+	cleanUpFn := func() {
+		errCleanUpStageOwner := s.Delete(context.Background(), stageRuntimeID)
+		if errCleanUpStageOwner != nil {
+			logr.WithError(errCleanUpStageOwner).Errorln("failed to delete failed stage owner")
+		}
+	}
+
+	if err := s.Create(ctx, &types.StageOwner{StageID: stageRuntimeID, PoolName: pool}); err != nil {
```
shubham149, line 104: we can update this line to first check if the entity is already present. If present, don't do anything. Same as we have done in destroy call. That will cover all the cases. No need to cleanup as it will happen in destroy anyways.

eoinmcafee00: The behaviour i've witness is if provisioning fails - the stage owner table isn't cleaned up. Yes the destroy call will clean it up now but that's not the behaviour the delegate is doing. Therefore this clean up necessary.

command/harness/setup.go
eoinmcafee00: Done

command/harness/destroy.go
```diff
@@ -38,13 +38,16 @@ func HandleDestroy(ctx context.Context, r *VMCleanupRequest, s store.StageOwnerS
 		WithField("stage_runtime_id", r.StageRuntimeID).
 		WithField("pool", r.PoolID)
 
-	if err := poolManager.Destroy(ctx, r.PoolID, inst.ID); err != nil {
+	if err = poolManager.Destroy(ctx, r.PoolID, inst.ID); err != nil {
```
vistaarjuneja, line 41: annot -> cannot

command/harness/destroy.go
```diff
@@ -38,13 +38,16 @@ func HandleDestroy(ctx context.Context, r *VMCleanupRequest, s store.StageOwnerS
...
 		return fmt.Errorf("annot destroy the instance: %w", err)
 	}
 	logr.Traceln("destroyed instance")
-
-	if err := s.Delete(ctx, r.StageRuntimeID); err != nil {
-		logrus.WithError(err).Errorln("failed to delete stage owner entity")
+	exists, _ := s.Find(ctx, r.StageRuntimeID, r.PoolID)
+	if exists != nil {
+		if deleteErr := s.Delete(ctx, r.StageRuntimeID); err != nil {
```
vistaarjuneja, line 47: err -> deleteErr

command/harness/destroy.go
```diff
@@ -38,13 +38,16 @@ func HandleDestroy(ctx context.Context, r *VMCleanupRequest, s store.StageOwnerS
...
+			logrus.WithError(deleteErr).Errorln("failed to delete stage owner entity")
```
vistaarjuneja, line 48: Use logr here since it has some other fields populated

command/harness/setup.go
```diff
@@ -96,17 +93,17 @@ func HandleSetup(ctx context.Context, r *SetupVMRequest, s store.StageOwnerStore
 	if !poolManager.Exists(pool) {
 		return nil, fmt.Errorf("pool not defined")
 	}
-
-	instance, err := poolManager.Provision(ctx, pool, env.Runner.Name, env.Settings.LiteEnginePath)
+	_, err := s.Find(ctx, stageRuntimeID, pool)
 	if err != nil {
-		return nil, fmt.Errorf("failed provisioning")
+		if err = s.Create(ctx, &types.StageOwner{StageID: stageRuntimeID, PoolName: pool}); err != nil {
```
vistaarjuneja, line 98: We have err on L97 as well as L98 (local variable). Not sure if that'll create a bug - perhaps we should change the error variable in the if condition?

command/harness/setup.go
```diff
@@ -96,17 +93,17 @@ func HandleSetup(ctx context.Context, r *SetupVMRequest, s store.StageOwnerStore
...
 	}
 
-	if instance.IsHibernated {
-		instance, err = poolManager.StartInstance(ctx, pool, instance.ID)
-		if err != nil {
-			return nil, fmt.Errorf("failed to start the instance up")
-		}
+	instance, err := poolManager.Provision(ctx, pool, env.Runner.Name, env.Settings.LiteEnginePath)
+	if err != nil {
+		logr.Errorln("failed to provision instance")
```
vistaarjuneja, line 105: Can we print the error here as well: `logr.WithError(err).Errorln("failed to provision instance")`

command/harness/destroy.go
eoinmcafee00: fixed

command/harness/destroy.go
eoinmcafee00: fixed

command/harness/destroy.go
eoinmcafee00: fixed

command/harness/setup.go
eoinmcafee00: I actually think this is fine?

vistaarjuneja (approved):

PR #156, opened 2022-08-12
(feat) Inject split test binary for Linux VMs (CIE)

rutvijmehta-harness:
## Summary
This PR downloads split test binary for Ubuntu Linux VMs where the driver is either delegate or dlite. It is downloaded from a GCS bucket where the binary is uploaded by the CIE Build Release pipelines. This is achieved by updating the cloudinit script for Ubuntu VMs which means the binary will be downloaded as part of VM startup. It also includes a minor refactor where the environment config is passed to the driver instead of individual parameters as suggested which includes updating a few structs as well.

## Local Test
Evidence for binary being present (downloaded) in the VM:
image

### DIFF
--- a/command/config/config.go
+++ b/command/config/config.go
@@ -279,14 +279,15 @@ type EnvConfig struct {
 	}
 
 	Settings struct {
-		LiteEnginePath string `envconfig:"DRONE_LITE_ENGINE_PATH" default:"https://github.com/harness/lite-engine/releases/download/v0.4.3/"`
-		DefaultDriver  string `envconfig:"DRONE_DEFAULT_DRIVER" default:"amazon"`
-		ReusePool      bool   `envconfig:"DRONE_REUSE_POOL" default:"false"`
-		BusyMaxAge     int64  `envconfig:"DRONE_SETTINGS_BUSY_MAX_AGE" default:"24"`
-		FreeMaxAge     int64  `envconfig:"DRONE_SETTINGS_FREE_MAX_AGE" default:"720"`
-		MinPoolSize    int    `envconfig:"DRONE_MIN_POOL_SIZE" default:"1"`
-		MaxPoolSize    int    `envconfig:"DRONE_MAX_POOL_SIZE" default:"2"`
-		EnableAutoPool bool   `envconfig:"DRONE_ENABLE_AUTO_POOL" default:"false"`
+		LiteEnginePath       string `envconfig:"DRONE_LITE_ENGINE_PATH" default:"https://github.com/harness/lite-engine/releases/download/v0.4.3/"`
+		DefaultDriver        string `envconfig:"DRONE_DEFAULT_DRIVER" default:"amazon"`
+		ReusePool            bool   `envconfig:"DRONE_REUSE_POOL" default:"false"`
+		BusyMaxAge           int64  `envconfig:"DRONE_SETTINGS_BUSY_MAX_AGE" default:"24"`
+		FreeMaxAge           int64  `envconfig:"DRONE_SETTINGS_FREE_MAX_AGE" default:"720"`
+		MinPoolSize          int    `envconfig:"DRONE_MIN_POOL_SIZE" default:"1"`
+		MaxPoolSize          int    `envconfig:"DRONE_MAX_POOL_SIZE" default:"2"`
+		EnableAutoPool       bool   `envconfig:"DRONE_ENABLE_AUTO_POOL" default:"false"`
+		HarnessTestBinaryURI string `envconfig:"DRONE_HARNESS_TEST_BINARY_URI"`
 	}
 
 	Server struct {
--- a/command/daemon/daemon.go
+++ b/command/daemon/daemon.go
@@ -100,7 +100,7 @@ func (c *daemonCommand) run(*kingpin.ParseContext) error {
 		),
 	)
 	store := database.ProvideInstanceStore(db)
-	poolManager := drivers.New(ctx, store, env.Settings.LiteEnginePath, env.Runner.Name)
+	poolManager := drivers.New(ctx, store, &env)
 
 	configPool, confErr := poolfile.ConfigPoolFile(c.poolFile, &env)
 	if confErr != nil {
--- a/command/harness/delegate/delegate.go
+++ b/command/harness/delegate/delegate.go
@@ -70,6 +70,9 @@ func (c *delegateCommand) run(*kingpin.ParseContext) error {
 	if err != nil {
 		return err
 	}
+	if env.Settings.HarnessTestBinaryURI == "" {
+		env.Settings.HarnessTestBinaryURI = "https://app.harness.io/storage/harness-download/harness-ti/split_tests"
+	}
 	c.env = env
 	// setup the global logrus logger.
 	harness.SetupLogger(&c.env)
@@ -89,7 +92,7 @@ func (c *delegateCommand) run(*kingpin.ParseContext) error {
 
 	instanceStore := database.ProvideInstanceStore(db)
 	c.stageOwnerStore = database.ProvideStageOwnerStore(db)
-	c.poolManager = drivers.New(ctx, instanceStore, c.env.Settings.LiteEnginePath, c.env.Runner.Name)
+	c.poolManager = drivers.New(ctx, instanceStore, &c.env)
 
 	_, err = harness.SetupPool(ctx, &c.env, c.poolManager, c.poolFile)
 	defer harness.Cleanup(&c.env, c.poolManager) //nolint: errcheck
--- a/command/harness/dlite/dlite.go
+++ b/command/harness/dlite/dlite.go
@@ -81,6 +81,9 @@ func (c *dliteCommand) run(*kingpin.ParseContext) error {
 	if err != nil {
 		return err
 	}
+	if env.Settings.HarnessTestBinaryURI == "" {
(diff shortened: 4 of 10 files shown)


internal/cloudinit/cloudinit.go
```diff
@@ -174,8 +175,12 @@ write_files:
   content: {{ .TLSKey | base64 }}
 runcmd:
 - 'ufw allow 9079'
-- 'wget "{{ .LiteEnginePath }}/lite-engine-{{ .Platform.OS }}-{{ .Platform.Arch }}" -O /usr/bin/lite-engine'
+- 'wget https://storage.googleapis.com/harness-ti/lite-engine -O /usr/bin/lite-engine'
```
rutvijmehta-harness, line 178: I'll update these URLs before merging.

internal/cloudinit/cloudinit.go
```diff
@@ -176,6 +177,10 @@ runcmd:
 - 'ufw allow 9079'
 - 'wget "{{ .LiteEnginePath }}/lite-engine-{{ .Platform.OS }}-{{ .Platform.Arch }}" -O /usr/bin/lite-engine'
 - 'chmod 777 /usr/bin/lite-engine'
+{{ if  (eq .IsHarness true) }}
+- 'wget https://storage.googleapis.com/harness-ti/split_tests -O /usr/bin/split_tests'
```
rutvijmehta-harness, line 181: I'll update this URL before merging.

command/harness/delegate/delegate.go
```diff
@@ -112,7 +112,7 @@ func (c *delegateCommand) run(*kingpin.ParseContext) error {
 
 	instanceStore := database.ProvideInstanceStore(db)
 	c.stageOwnerStore = database.NewStageOwnerStore(db)
-	c.poolManager = drivers.New(ctx, instanceStore, c.env.Settings.LiteEnginePath, c.env.Runner.Name)
+	c.poolManager = drivers.New(ctx, instanceStore, c.env.Settings.LiteEnginePath, c.env.Runner.Name, false)
```
shubham149, line 115: we need to set it to true

command/harness/dlite/dlite.go
```diff
@@ -105,7 +105,7 @@ func (c *dliteCommand) run(*kingpin.ParseContext) error {
 
 	instanceStore := database.ProvideInstanceStore(db)
 	c.stageOwnerStore = database.NewStageOwnerStore(db)
-	c.poolManager = drivers.New(ctx, instanceStore, c.env.Settings.LiteEnginePath, c.env.Runner.Name)
+	c.poolManager = drivers.New(ctx, instanceStore, c.env.Settings.LiteEnginePath, c.env.Runner.Name, false)
```
shubham149, line 108: ditto

command/daemon/daemon.go
```diff
@@ -100,7 +100,7 @@ func (c *daemonCommand) run(*kingpin.ParseContext) error {
 		),
 	)
 	store := database.ProvideInstanceStore(db)
-	poolManager := drivers.New(ctx, store, env.Settings.LiteEnginePath, env.Runner.Name)
+	poolManager := drivers.New(ctx, store, env.Settings.LiteEnginePath, env.Runner.Name, false)
```
eoinmcafee00, line 103: This should be an environment variable and injected in through the config struct. I have no context of what false is from looking at this. It would also mean you are touching less places in code as the config struct in injected up the layers anyways.

eoinmcafee00:
Hey @rutvijmehta-harness 

I don't see any e2e test evidence that this works. Can you provide a screenshot or logs that the change in fact works as expected please?

command/daemon/daemon.go
rutvijmehta-harness: That was my initial approach where I added IsHarness to EnvConfig and set config.IsHarness = true for CIE and not set it for others. That also required adding IsHarness to execCommand in exec.go and not set it. Please let me know if that approach works. I'll upload a new patch.

command/harness/delegate/delegate.go
rutvijmehta-harness: Made this typo when copying from drone-runner-aws repo to its fork in my local. I'll fix it.

command/harness/dlite/dlite.go
rutvijmehta-harness: Ack.

rutvijmehta-harness:
image
@eoinmcafee00 addressed Shubham's comment and ran the build locally. Happy to jump on a call to explain the end to end working.

tphoney (changes requested):
Hey @rutvijmehta-harness there are a number of things that should be changed. 

1. give isHarness a better name. technically everything is harness.
2. add a comment to what you are doing in the cloudinit file. Is this downloading pictures of cats ?
3. why do we not have this for other operating systems ?
4. this only affects one file, it should be set as an environment variable as @eoinmcafee00 mentioned. it should not change the method signatures for daemon / exec / delegate / setup dlite

rutvijmehta-harness:
Thanks @tphoney for the review.
1. Do you think isCIE better?
2. Ack.
3. Once the implementation is finalized on how we determine CIE/Drone, I'll add it to other operation systems and also test it.
4. I'd like to discuss this approach. I want to pass a bool in drivers.New and I am not sure how environment variables would help without changing the method signatures for daemon, exec, etc.

eoinmcafee00:
@rutvijmehta-harness 
What about a slight refactor of the method signature so that we pass the config struct into the driver - that should get rid of needing to pass runner name & lite engine path. Then in the manager where we do pool.driver.create, extend the create option struct and pass in the env variable you've created. 

isCIE isn't great either, it should describe what it's doing "InstallHarnessTestBinaray"?

rutvijmehta-harness:
@tphoney @eoinmcafee00 I have addressed the above comments.

internal/cloudinit/cloudinit.go
```diff
@@ -20,11 +20,12 @@ import (
...
-	CACert         string
-	TLSCert        string
-	TLSKey         string
-	Platform       types.Platform
+	LiteEnginePath        string
+	CACert                string
+	TLSCert               string
+	TLSKey                string
+	Platform              types.Platform
+	DownloadHarnessBinary bool
```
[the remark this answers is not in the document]
rutvijmehta-harness: @tphoney do you want this string to be constructed in here or construct it in delegate and pass it here?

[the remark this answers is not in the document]
tphoney: either way, what ever suits. 
it is to prevent the need of rebuilding the binary if the download location changes.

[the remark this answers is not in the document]
rutvijmehta-harness: Done. I set `https://app.harness.io/storage/harness-download/harness-ti/split_tests` as `HarnessDownloadURI` which is then used in the cloudinit script. This config can be overridden by env variable. I couldn't add arch and os in delegate/dlite code as I don't have access to those values at that time.

command/config/config.go
```diff
@@ -279,14 +279,15 @@ type EnvConfig struct {
...
-		EnableAutoPool bool   `envconfig:"DRONE_ENABLE_AUTO_POOL" default:"false"`
+		LiteEnginePath     string `envconfig:"DRONE_LITE_ENGINE_PATH" default:"https://github.com/harness/lite-engine/releases/download/v0.4.3/"`
+		DefaultDriver      string `envconfig:"DRONE_DEFAULT_DRIVER" default:"amazon"`
+		ReusePool          bool   `envconfig:"DRONE_REUSE_POOL" default:"false"`
+		BusyMaxAge         int64  `envconfig:"DRONE_SETTINGS_BUSY_MAX_AGE" default:"24"`
+		FreeMaxAge         int64  `envconfig:"DRONE_SETTINGS_FREE_MAX_AGE" default:"720"`
+		MinPoolSize        int    `envconfig:"DRONE_MIN_POOL_SIZE" default:"1"`
+		MaxPoolSize        int    `envconfig:"DRONE_MAX_POOL_SIZE" default:"2"`
+		EnableAutoPool     bool   `envconfig:"DRONE_ENABLE_AUTO_POOL" default:"false"`
+		HarnessDownloadURI string `envconfig:"DRONE_HARNESS_DOWNLOAD_URI"`
```
eoinmcafee00, line 290: Set a default similar to how we do it with lite-engine

command/harness/delegate/delegate.go
```diff
@@ -70,6 +70,9 @@ func (c *delegateCommand) run(*kingpin.ParseContext) error {
 	if err != nil {
 		return err
 	}
+	if env.Settings.HarnessDownloadURI == "" {
```
eoinmcafee00, line 73: not needed if you set the default

command/config/config.go
eoinmcafee00: Also, HarnessDownloadURI tells me nothing. what's it downloading? 'HarnessTestBinaryURI'

rutvijmehta-harness: Ack. Will replace HarnessDownloadURI with HarnessTestBinaryURI

command/harness/delegate/delegate.go
rutvijmehta-harness: I explicitly didn't set default as we only want this value to be set for delegate and dlite. If I'll set a default, it'll be present for other runners as well which we don't want

command/config/config.go
rutvijmehta-harness: @eoinmcafee00 done

eoinmcafee00 (approved):

tphoney (approved):

PR #164, opened 2022-08-26
Add tags to a build instance

shubham149:
# Commit Checklist

## The Basics

### DIFF
--- a/internal/drivers/amazon/driver.go
+++ b/internal/drivers/amazon/driver.go
@@ -384,6 +384,21 @@ func (p *config) Logs(ctx context.Context, instanceID string) (string, error) {
 	return string(decoded), nil
 }
 
+func (p *config) SetTags(ctx context.Context, instanceID string,
+	tags map[string]string) error {
+	in := &ec2.CreateTagsInput{
+		Resources: []*string{aws.String(instanceID)},
+	}
+	for key, value := range tags {
+		in.Tags = append(in.Tags, &ec2.Tag{
+			Key:   aws.String(key),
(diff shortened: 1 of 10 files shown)


internal/drivers/google/driver.go
```diff
@@ -244,6 +244,26 @@ func (p *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (in
 	return &instanceMap, nil
 }
 
+func (p *config) SetTags(ctx context.Context, instanceID string, tags map[string]string) error {
+	logr := logger.FromContext(ctx).
```
vistaarjuneja, line 248: Can we return immediately if tags are empty? We don't need to do a get if there are no tags

internal/drivers/amazon/driver.go
```diff
@@ -384,6 +384,21 @@ func (p *config) Logs(ctx context.Context, instanceID string) (string, error) {
 	return string(decoded), nil
 }
 
+func (p *config) SetTags(ctx context.Context, instanceID string,
+	tags map[string]string) error {
+	in := &ec2.CreateTagsInput{
```
vistaarjuneja, line 389: Likewise, if tags are empty we should return immediately

shubham149: Added the check in manager.go

internal/drivers/google/driver.go
shubham149: Added the check in manager.go

vistaarjuneja (approved):

PR #165, opened 2022-08-27
(feat) add maintenance support in dlite mode

vistaarjuneja:
# Commit Checklist

## The Basics


command/harness/delegate/delegate.go
```diff
@@ -118,6 +95,7 @@ func (c *delegateCommand) run(*kingpin.ParseContext) error {
 	if err != nil {
 		return err
 	}
+	defer harness.Cleanup(&c.env, c.poolManager)
```
shubham149, line 98: We can retain using a separate go routine for cleanup instead of defer. Defer will run when the main has exited. I have seen sometimes the exit gets stuck and program never exits. In that case, cleanup will never get called.

command/harness/dlite/dlite.go
```diff
@@ -109,25 +106,46 @@ func (c *dliteCommand) run(*kingpin.ParseContext) error {
 
 	poolConfig, err := harness.SetupPool(ctx, &c.env, c.poolManager, c.poolFile)
 	if err != nil {
+		logrus.WithError(err).Error("could not setup pool")
 		return err
 	}
+	defer harness.Cleanup(&c.env, c.poolManager)
```
shubham149, line 112: Ditto

command/harness/dlite/handler.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/go-chi/chi"
+	"github.com/wings-software/dlite/client"
+	"github.com/wings-software/dlite/poller"
+)
+
+func Handler(p *poller.Poller) http.Handler {
+	r := chi.NewRouter()
+	r.Use(harness.Middleware)
+
+	r.Mount("/maintenance_mode", func() http.Handler {
```
shubham149, line 17: Add get on maintenance mode to print the current state whether it is in PROD mode or maintenance mode

command/harness/dlite/handler.go
```diff
@@ -0,0 +1,43 @@
...
+	"io"
+	"net/http"
+
+	"github.com/drone-runners/drone-runner-aws/command/harness"
+	"github.com/go-chi/chi"
+	"github.com/wings-software/dlite/client"
+	"github.com/wings-software/dlite/poller"
+)
+
+func Handler(p *poller.Poller) http.Handler {
```
shubham149, line 13: Can we move the http server code to dlite? Dlite can add a module that you can use to start the server. This handler will be present in dlite as well. Server code starter can take an interface 

type handler interface {
   EnableMaintenanceMode
   DisableMaintenanceMode
}

Everyone using dlite can implement this interface and start the server and use it.

command/harness/delegate/delegate.go
vistaarjuneja: Done. I kept the defer as well since for example if there are any VMs left over after `SetupPool` and it returns in an error, then we still need to delete the VMs.

command/harness/dlite/handler.go
vistaarjuneja: I thought about this initially but I would prefer to keep this here until we have a more generic abstraction for what maintenance mode means in dlite. I'm not sure yet how well this would translate to other runners. For now I have kept the filter logic in dlite which anyone can use and will move the maintenance mode logic to dlite as well once it's more generic.

command/harness/delegate/delegate.go
shubham149: Move cleanup to L96 in that case.

vistaarjuneja: Done

shubham149 (approved):

PR #170, opened 2022-09-09
Add curl tool to the docker image

shubham149:
Curl is required for enabling/disabling maintenance mode on dlite running in a kubernetes cluster for hosted vms.

# Commit Checklist

## The Basics

### DIFF
--- a/docker/Dockerfile.linux.amd64
+++ b/docker/Dockerfile.linux.amd64
@@ -2,6 +2,7 @@ FROM alpine:3.6 as alpine
 RUN apk add -U --no-cache ca-certificates
 
 FROM alpine:3.6
+RUN apk add -U --no-cache curl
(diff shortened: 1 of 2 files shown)


tphoney:
just for clarification, this will be for an engineer to run this as part of a diagnosis ?

shubham149:
@tphoney This will be used as part of automatic upgrades of runner for hosted VMs. Runner would be running as a kubernetes deployment with maintenance mode api exposed via http but it is not exposed as k8 service. We will curl on the K8 pods and call curl to enable the maintenance mode as part of upgrade process.

shubham149:
Also, it will be used by engineers to check whether the maintenance mode is enabled on runner or not during debugging.

tphoney (approved):

PR #179, opened 2022-09-19
Reduce deadlock time in vm setup

shubham149:
In case of burst of vm setup request e.g. 1000 in a minute, a lot of setup calls were in waiting state for lock for more than 15 minutes. 

# Commit Checklist

## The Basics

### DIFF
--- a/internal/drivers/manager.go
+++ b/internal/drivers/manager.go
@@ -274,20 +274,21 @@ func (m *Manager) Provision(ctx context.Context, poolName, serverName, liteEngin
 		return nil, fmt.Errorf("provision: pool name %q not found", poolName)
 	}
 
-	pool.Lock()
-	defer pool.Unlock()
-
 	strategy := m.strategy
 	if strategy == nil {
 		strategy = Greedy{}
 	}
 
+	pool.Lock()
+
 	busy, free, err := m.List(ctx, pool)
 	if err != nil {
+		pool.Unlock()
 		return nil, fmt.Errorf("provision: failed to list instances of %q pool: %w", poolName, err)
 	}
 
 	if len(free) == 0 {
+		pool.Unlock()
 		if canCreate := strategy.CanCreate(pool.MinSize, pool.MaxSize, len(busy), len(free)); !canCreate {
 			return nil, ErrorNoInstanceAvailable
 		}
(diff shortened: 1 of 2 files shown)


internal/drivers/manager.go
```diff
@@ -309,13 +310,15 @@ func (m *Manager) Provision(ctx context.Context, poolName, serverName, liteEngin
...
+		pool.Unlock()
 		return nil, fmt.Errorf("provision: failed to tag an instance in %q pool: %w", poolName, err)
 	}
+	pool.Unlock()
 
 	// the go routine here uses the global context because this function is called
 	// from setup API call (and we can't use HTTP request context for async tasks)
 	go func(ctx context.Context) {
-		_ = m.buildPoolWithMutex(ctx, pool)
+		_, _ = m.setupInstance(ctx, pool, false)
```
vistaarjuneja, line 321: This is changing the behavior a little bit: previously, for each init call, looks like the pool would be rebuilt again and any extra instances removed. Now we only setup a single instance. Is there some other background thread which is taking care of maintaining the pool?

shubham149: instance purger takes care of it. Ideally deletion from pool is not required at the time of provisioning request

vistaarjuneja: This is creating instances as well right? Let's say there are x instances in the pool and we get x parallel requests. They will each pick up a warm instance. After that it seems creation is happening only one by one with the newer code?

shubham149: that will happen in parallel since new vm creation is via goroutines

shubham149: Tested with scale testing of 1000 parallel builds and no issues were observed.

tphoney: Nice job testing it, how long did you let the runner run for. Did you observe the behaviour of the pools over a longer period (1hr) ?

vistaarjuneja: @tphoney we did this as part of scale testing for hosted VMs. It was triggering ~100 builds/minute for around 30 minutes. Once all the builds were done, we saw that the pool count was back to the pool size.

With this PR, we're either creating a VM from the pool (if it doesn't exist). Or we use an instance from the pool and create another instance at the same time. The pool is created initially and thus the count is maintained with the completion of all the builds as well.

This works for the greedy strategy (which is currently being used) but it would not maintain the minMax strategy since currently the pool gets maintained on every call (and extra VMs removed). If we're not planning to support that (and start honoring limit in the pool yaml), this behaviour should be fine.

eoinmcafee00 (approved):

PR #189, opened 2022-10-30
Changes in cloud init script for installing tmate

HenAmar:
# Commit Checklist

## The Basics


internal/cloudinit/cloudinit.go
```diff
@@ -201,7 +201,14 @@ runcmd:
 {{ end }}
 - 'touch /root/.env'
 - '[ -f "/etc/environment" ] && cp "/etc/environment" /root/.env'
-- '/usr/bin/lite-engine server --env-file /root/.env > /var/log/lite-engine.log 2>&1 &'`
+- '/usr/bin/lite-engine server --env-file /root/.env > /var/log/lite-engine.log 2>&1 &'
+- 'mkdir /addon'
```
shubham149, line 205: Addon volume is not used in vms. It is used specifically in k8 only.

shubham149: Also, this needs to work for both linux amd64 and linux arm64

internal/cloudinit/cloudinit.go
```diff
@@ -201,7 +201,14 @@ runcmd:
...
+- 'wget https://github.com/tmate-io/tmate/releases/download/2.4.0/tmate-2.4.0-static-linux-amd64.tar.xz -O /addon/tmate.xz'
```
shubham149, line 206: This needs to work for both linux amd64 and linux arm64

tphoney (changes requested):
There are a couple of things: 

1. you have not implemented this for the amazon linux build.
2. most importantly there is no way to disable this functionality. This installation block should be wrapped in a conditional. Ideally we should be able to disable/enable this with an environment variable. 
3. If we are enabling this via cloud init, as seen. Then we should remove the alternate way of installing it for drone. 
4. Are there binaries for mac ?

HenAmar:
Hey @tphoney :

> 1. you have not implemented this for the amazon linux build. 
Added

> 2. most importantly there is no way to disable this functionality. This installation block should be wrapped in a conditional. Ideally we should be able to disable/enable this with an environment variable.
I took advantage of the existing "DRONE_TMATE_ENABLED" var and populate it all the way to use it. 

> 3. If we are enabling this via cloud init, as seen. Then we should remove the alternate way of installing it for drone.
I put the foundation for it and it should be easy to do it now, but I would like it to be done in a different scope and not part of this PR

> 4. Are there binaries for mac ?
No.

tphoney (approved):

eoinmcafee00 (approved):

PR #192, opened 2022-11-04
Add plugin binary to cloudinit to support containerless plugin, action and bitrise steps

shubham149:
# Commit Checklist

## The Basics

### DIFF
--- a/command/config/config.go
+++ b/command/config/config.go
@@ -288,6 +288,7 @@ type EnvConfig struct {
 		MaxPoolSize          int    `envconfig:"DRONE_MAX_POOL_SIZE" default:"2"`
 		EnableAutoPool       bool   `envconfig:"DRONE_ENABLE_AUTO_POOL" default:"false"`
 		HarnessTestBinaryURI string `envconfig:"DRONE_HARNESS_TEST_BINARY_URI"`
+		PluginBinaryURI      string `envconfig:"DRONE_PLUGIN_BINARY_URI" default:"https://github.com/drone/plugin/releases/download/v0.1.2-alpha"`
 	}
 
 	Server struct {
--- a/internal/cloudinit/cloudinit.go
+++ b/internal/cloudinit/cloudinit.go
@@ -26,6 +26,7 @@ type Params struct {
 	TLSKey               string
 	Platform             types.Platform
 	HarnessTestBinaryURI string
+	PluginBinaryURI      string
 }
 
 var funcs = map[string]interface{}{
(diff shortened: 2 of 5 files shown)


tphoney (commented):
Does this mean there is no windows or mac support ?

command/config/config.go
```diff
@@ -288,6 +288,7 @@ type EnvConfig struct {
 		MaxPoolSize          int    `envconfig:"DRONE_MAX_POOL_SIZE" default:"2"`
 		EnableAutoPool       bool   `envconfig:"DRONE_ENABLE_AUTO_POOL" default:"false"`
 		HarnessTestBinaryURI string `envconfig:"DRONE_HARNESS_TEST_BINARY_URI"`
+		PluginBinaryURI      string `envconfig:"DRONE_PLUGIN_BINARY_URI" default:"https://github.com/drone/plugin/releases/download/v0.1.2-alpha"`
```
bradrydzewski, line 291: @shubham149 what do you think about linking to the latest binary by default for now?  If we can keep the command line flags stable, it means we can iterate without having to perform new lite engine releases?

> To link directly to a download of your latest release asset that was manually uploaded, link to /owner/name/releases/latest/download/asset-name.zip.

shubham149: There are a lot of possibilities for introducing a regression in anything that is using plugin binary if we use latest. Using a versioned binary reduces those issues. We will have automated tests that run as part of new releases in harness ci which will catch regressions. But that won't be possible with latest version usage.

We are following the same convention for lite-engine in runner and it has worked well for us.

shubham149:
@tphoney Plugin binary supports windows as well as mac. We will need to update the cloudinit file. We can add its support in runner in next phase.

command/config/config.go
bradrydzewski: @shubham149 in Drone many of our customers use the `:latest` version of plugin.  It requires discipline to avoid breaking changes, but it means customers always get the latest and greatest.  It has worked well for Drone over the years.  It is ok if we are not ready to pull the `latest` release, but let's make this a goal for the future. Discipline and backward compatibility is a huge reason Drone sold over the years;  enterprises value stability over almost everything else.

shubham149: We can use versioned approach for now. Right now, we have a weekly release cadence for harness CI which should be good enough for most of the cases. 

I am worried about engineering team spending more time debugging the customer issues related to plugin in case we use latest version. We will need to add tests covering all the cases in plugin repo drone builds before we can move to latest release. Also, we will have to setup mechanisms for debugging customer issues with latest versioned approach.

tphoney: To add to Brads point, will the default value ever be used ? 
if it is always used  then there will need to be regular weekly releases of this runner. Which means that any new large changes going in, have to be introduced in a week. 
If this value is always overridden then using a default value of latest seems to be natural.

tphoney:
@eoinmcafee00 

> @tphoney Plugin binary supports windows as well as mac. We will need to update the cloudinit file. We can add its support in runner in next phase.

internal/cloudinit/cloudinit.go
```diff
@@ -182,6 +183,10 @@ runcmd:
 - 'wget "{{ .HarnessTestBinaryURI }}/{{ .Platform.Arch }}/{{ .Platform.OS }}/bin/split_tests-{{ .Platform.OS }}_{{ .Platform.Arch }}" -O /usr/bin/split_tests'
 - 'chmod 777 /usr/bin/split_tests'
 {{ end }}
+{{ if .PluginBinaryURI }}
+- 'wget {{ .PluginBinaryURI }}/plugin-{{ .Platform.OS }}-{{ .Platform.Arch }}  -O /usr/bin/plugin'
+- 'chmod 777 /usr/bin/plugin'
+{{ end }}
```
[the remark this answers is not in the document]
shubham149: I will add support for windows and mac in next set of PR.

tphoney (approved):
Happy for the change to go in knowing:

1. mac support / other os support is incoming
2. the default value will change to latest once we settle down in the plugin development.

PR #207, opened 2022-12-05
(bug) add nil check if pool not found

eoinmcafee00:


internal/drivers/manager.go
```diff
@@ -98,6 +98,9 @@ func (m *Manager) GetInstanceByStageID(ctx context.Context, poolName, stage stri
 	}
 
 	pool := m.poolMap[poolName]
+	if pool == nil {
+		return nil, fmt.Errorf("GetInstanceByStageID: pool name %q not found", poolName)
```
vistaarjuneja, line 102: Let's log the error message, also add stage_runtime_id like L107.

eoinmcafee00: updated

vistaarjuneja (approved):

PR #208, opened 2022-12-06
(bug) - fixes scheduling timeout issue on ankabuild

eoinmcafee00:
Fixes issue where if an instance fails to schedule on a node, it will remove it from the queue.

### DIFF
--- a/internal/drivers/ankabuild/driver.go
+++ b/internal/drivers/ankabuild/driver.go
@@ -78,10 +78,15 @@ func (c *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (in
 	}
 	var id = response.Body[0]
 	vm := &vmResponse{}
-	for i := 1; i <= 50000; i++ {
+	for i := 1; i <= 10000; i++ {
 		vm, err = c.ankaClient.VMFind(ctx, id)
 		if err != nil {
 			return nil, err
+		}
+		if vm.Body.InstanceState == "Scheduling" {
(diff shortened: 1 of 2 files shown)


internal/drivers/ankabuild/driver.go
```diff
@@ -78,24 +78,36 @@ func (c *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (in
 	}
 	var id = response.Body[0]
 	vm := &vmResponse{}
-	for i := 1; i <= 50000; i++ {
+	for i := 1; i <= 100; i++ {
```
vistaarjuneja, line 81: For future: might be better to use exponential backoff here instead of a for loop. Since we don't know what is the max possible time it can stay in the for loop before going to L103. (100 * 5 seconds?)

tphoney, line 81: there are other ways to achieve this with a context, i used something similar to what brad used for ssh connections here https://github.com/drone-runners/drone-runner-aws/pull/6/files or here https://golang.cafe/blog/golang-context-with-timeout-example.html
it removes the reliance on sleeps. 
i do understand the urgency of this though, but i can see the sleeps / timing being an issue again if there are networking issues.

eoinmcafee00: Yeah lets revisit in future;

eoinmcafee00: Agreed - I think for now we should come back to making this change.

vistaarjuneja (approved):

PR #209, opened 2022-12-06
fixes issue with timestamp being used on machine name & allow session…

eoinmcafee00:
Fixes:
1. Pass in optional session token for AWS creds
2. Remove timestamp from instance name creation - issues if lots of VMS spun up at once. Name has to be unique.

### DIFF
--- a/internal/drivers/amazon/driver.go
+++ b/internal/drivers/amazon/driver.go
@@ -20,6 +20,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/cenkalti/backoff/v4"
+	"github.com/dchest/uniuri"
 )
 
 // config is a struct that implements drivers.Pool interface
@@ -31,6 +32,7 @@ type config struct {
 
 	accessKeyID     string
 	secretAccessKey string
+	sessionToken    string
 	keyPairName     string
 
 	rootDir string
@@ -67,7 +69,11 @@ func New(opts ...Option) (drivers.Driver, error) {
 			MaxRetries: aws.Int(p.retries),
 		}
 		if p.accessKeyID != "" && p.secretAccessKey != "" {
-			config.Credentials = credentials.NewStaticCredentials(p.accessKeyID, p.secretAccessKey, "")
+			if p.sessionToken != "" {
+				config.Credentials = credentials.NewStaticCredentials(p.accessKeyID, p.secretAccessKey, p.sessionToken)
+			} else {
+				config.Credentials = credentials.NewStaticCredentials(p.accessKeyID, p.secretAccessKey, "")
+			}
 		}
 		mySession := session.Must(session.NewSession())
 		p.service = ec2.New(mySession, config)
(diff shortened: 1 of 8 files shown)


internal/drivers/digitalocean/driver.go
```diff
@@ -69,7 +70,7 @@ func (p *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (in
 		WithField("pool", opts.PoolName).
 		WithField("image", p.image).
 		WithField("hibernate", p.CanHibernate())
-	var name = fmt.Sprintf(opts.RunnerName+"-"+opts.PoolName+"-%d", startTime.Unix())
+	var name = fmt.Sprintf(opts.RunnerName+"-"+opts.PoolName+"-%d", uniuri.NewLen(8)) //nolint
```
vistaarjuneja, line 73: Nit: will %d work for this as it would be a string?

internal/drivers/anka/driver.go
```diff
@@ -68,7 +68,7 @@ func (p *config) CanHibernate() bool {
 func (p *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (instance *types.Instance, err error) {
 	startTime := time.Now()
 	uData := lehelper.GenerateUserdata(p.userData, opts)
-	machineName := fmt.Sprintf(opts.RunnerName+"-"+"-%d", rand.Int()) //nolint
+	machineName := fmt.Sprintf(opts.RunnerName + uniuri.NewLen(8)) //nolint
```
vistaarjuneja, line 71: Maybe keep the "-" as it helps create separation

internal/drivers/amazon/driver.go
```diff
@@ -186,7 +192,7 @@ func (p *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (in
 		WithField("image", p.image).
 		WithField("size", p.size).
 		WithField("hibernate", p.CanHibernate())
-	var name = fmt.Sprintf(opts.RunnerName+"-"+opts.PoolName+"-%d", startTime.Unix())
+	var name = fmt.Sprintf(opts.RunnerName + uniuri.NewLen(8)) //nolint
```
vistaarjuneja, line 195: Can we keep the pool name as well? 
`<runnerName>-<poolName>-<randomString>`

internal/drivers/anka/driver.go
tphoney, line 71: why are we disabling lint for this ?

eoinmcafee00: updated to //nolint:gomnd

internal/drivers/digitalocean/driver.go
eoinmcafee00: Fixed to %s

tphoney (approved):

internal/drivers/amazon/driver.go
```diff
@@ -67,7 +69,11 @@ func New(opts ...Option) (drivers.Driver, error) {
 			MaxRetries: aws.Int(p.retries),
 		}
 		if p.accessKeyID != "" && p.secretAccessKey != "" {
-			config.Credentials = credentials.NewStaticCredentials(p.accessKeyID, p.secretAccessKey, "")
+			if p.sessionToken != "" {
```
vistaarjuneja, line 72: Small nit: no need for this check. Just pass in p.sessionToken

vistaarjuneja (approved):

PR #222, opened 2023-01-05
Add nomad driver for bare metal integration

vistaarjuneja:
Basic flow:

1. The Create call goes to the nomad server - which registers the job, finds an appropriate node for it, and executes it. Once the job has executed successfully, we store the node information in the DB and return back. This is used to route the destroy job to the correct node.
2. The Create call creates a VM using ignite (which uses firecracker underneath). It also enables port forwarding and starts up the lite engine inside the VM.
3. All the interactions with the lite engine happen directly via REST on the node with the forwarded port, with required certificates (similar to other drivers)
4. Once the execution is done, the destroy task gets routed to the same node where the create happened and it kills the VM.


internal/poolfile/config.go
```diff
@@ -32,6 +33,7 @@ func ProcessPool(poolFile *config.PoolFile, runnerName string) ([]drivers.Pool,
 
 	for i := range poolFile.Instances {
 		instance := poolFile.Instances[i]
+		fmt.Printf("instance type is: %s", instance.Type)
```
shubham149, line 36: remove this line

internal/drivers/nomad/driver.go
```diff
@@ -0,0 +1,395 @@
...
+			break L
+		default:
+			q := &api.QueryOptions{WaitTime: 15 * time.Second, WaitIndex: waitIndex}
+			var qm *api.QueryMeta
+			// Get the job status
+			job, qm, err = client.Jobs().Info(id, q)
+			if job == nil {
+				continue
+			}
+			if err != nil {
```
shubham149, line 345: Shouldn't err be handled before job==nil code

eoinmcafee00 (approved):

internal/drivers/nomad/driver.go
```diff
@@ -0,0 +1,443 @@
...
+	vm := strings.ToLower(random(20)) //nolint:gomnd
+
+	hostPath := fmt.Sprintf("/usr/local/bin/%s.sh", vm)
+	vmPath := fmt.Sprintf("/usr/bin/%s.sh", vm)
+	jobID := fmt.Sprintf("init_job_%s", vm)
+	port := fmt.Sprintf("NOMAD_PORT_%s", vm)
+
+	logr := logger.FromContext(ctx).WithField("vm", vm).WithField("job_id", jobID)
+
+	runCmd := fmt.Sprintf("%s run %s --name %s --cpus %s --memory %s --size 50GB --ssh --runtime=docker --ports $%s:%s --copy-files %s:%s",
```
harshjain12, line 111: Do not hard code size

PR #226, opened 2023-01-11
Reduce locking scope for hibernate call

shubham149:
# Commit Checklist

## The Basics


internal/drivers/manager.go
```diff
@@ -38,6 +38,9 @@ type (
 	poolEntry struct {
 		sync.Mutex
 		Pool
+
+		LockByInstance map[string]*sync.Mutex
+		InstLock       sync.Mutex
```
vistaarjuneja, line 43: why do we have both a lock by instance map and an InstLock; and another mutex?

shubham149: InstLock is to ensure that fetching instance lock happens over a lock

shubham149:
Found a better way to handle it

PR #234, opened 2023-01-18
(feat) add retry and refactor creation step logic

eoinmcafee00:


internal/drivers/ankabuild/driver.go
```diff
@@ -104,8 +106,15 @@ func (c *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (in
 		deleteErr := c.ankaClient.VMDelete(ctx, vm.Body.InstanceID)
 		if deleteErr != nil {
 			logrus.Errorf("ankabuild: error deleting vm: %s", deleteErr)
+			return nil, deleteErr
+		}
+		// VMs creation can randomly fail, retry 3 times before returning error
+		if retry < 3 {
+			logrus.Infof("ankabuild: retry vm creation attempt: %d", retry)
+			retry++
```
vistaarjuneja, line 114: wouldn't this behaviour be different from expected? You have retry as a global variable so anytime Create is called it will increment retry by 1 - for each Create() from any instance of config.
Also, it might be better to retry the specific API call which needs to be retried as opposed to calling the function again

eoinmcafee00: Ready for another review

internal/drivers/ankabuild/driver.go
```diff
@@ -138,6 +107,73 @@ func (c *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (in
...
+				time.Sleep(retryInterval)
+				retryInterval = retryInterval * 2
+				continue
+			} else {
+				return nil, fmt.Errorf("failed to create vm after %d retries: %v", maxRetries, err)
+			}
+		}
+		var id = response.Body[0]
+		vm = &vmResponse{}
+		vm, _ = c.FindVM(ctx, id, retryInterval)
```
vistaarjuneja, line 128: This logic is not fully clear to me. So `CreateVM` is taking a `retryInterval` which seems to be a backoff interval and doubles on each retry - but why is this passed into the `FindVM` function - which is treating the `retryInterval` as max time it will retry?

eoinmcafee00: I basically just reused the retry Interval rather than creating 2 distinct ones 'retryIntervalCreate & retryIntervalfind'

eoinmcafee00: in the case of find vm it's used for the maximum time that the function will keep trying

eoinmcafee00: I've separated these out to stop confusion

internal/drivers/manager.go
```diff
@@ -521,6 +521,13 @@ func (m *Manager) setupInstance(ctx context.Context, pool *poolEntry, inuse bool
 			Errorln("manager: failed to generate certificates")
 		return nil, err
 	}
+	maxRetries := 3
```
vistaarjuneja, line 524: I feel this should be set at the driver layer instead of here -  mainly because each driver might have different values or retry logic it wants to use

eoinmcafee00: done

vistaarjuneja (approved):

PR #239, opened 2023-02-01
(feat) check instance still exists before provisioning it.

eoinmcafee00:


internal/drivers/ankabuild/driver.go
```diff
@@ -242,3 +244,14 @@ func (c *config) DriverName() string {
 func (c *config) CanHibernate() bool {
 	return false
 }
+
+func (c *config) InstanceExists(ctx context.Context, instanceID string) (bool, error) {
+	instance, err := c.ankaClient.VMFind(ctx, instanceID)
+	if instance == nil || err != nil {
+		return false, nil
```
vistaarjuneja, line 251: Can we add a log message either here or in the manager level

internal/drivers/manager.go
```diff
@@ -327,6 +327,12 @@ func (m *Manager) Provision(ctx context.Context, poolName, serverName string, en
 	})
 
 	inst := free[0]
+	exists, _ := pool.Driver.InstanceExists(ctx, inst.ID)
+	if !exists {
+		pool.Unlock()
+		_ = m.instanceStore.Delete(ctx, inst.ID)
+		return m.Provision(ctx, poolName, serverName, env)
```
vistaarjuneja, line 334: Discussed offline regarding the recursive nature of this

eoinmcafee00: done

PR #253, opened 2023-03-14
add retries on Harness destroy task

vistaarjuneja:
If a destroy task comes while an initialize task is ongoing, destroy was not getting processed sometimes. This PR adds exponential backoff retries to the destroy task so that it completes.

### DIFF
--- a/command/harness/destroy.go
+++ b/command/harness/destroy.go
@@ -3,12 +3,18 @@ package harness
 import (
 	"context"
 	"fmt"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/drone-runners/drone-runner-aws/internal/drivers"
 	ierrors "github.com/drone-runners/drone-runner-aws/internal/types"
 	"github.com/drone-runners/drone-runner-aws/store"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+)
+
+var (
+	destroyTimeout = 10 * time.Minute
 )
 
 type VMCleanupRequest struct {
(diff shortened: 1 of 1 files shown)


command/harness/destroy.go
```diff
@@ -3,14 +3,20 @@ package harness
...
+	"github.com/cenkalti/backoff/v4"
 	"github.com/drone-runners/drone-runner-aws/internal/drivers"
 	ierrors "github.com/drone-runners/drone-runner-aws/internal/types"
 	"github.com/drone-runners/drone-runner-aws/store"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
+var (
+	destroyTimeout = 10 * time.Minute
```
eoinmcafee00, line 17: I think it would be better to pass the value in, than declared at the top? Also is the value to high?

vistaarjuneja: This value is the max time an init task can be stuck. Currently the values are defined in the runner since it's doing the retries itself instead of being retried from the manager framework: https://github.com/drone-runners/drone-runner-aws/blob/master/command/harness/dlite/init.go#L16

eoinmcafee00 (approved):

command/harness/destroy.go
```diff
@@ -3,14 +3,20 @@ package harness
 import (
 	"context"
 	"fmt"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
```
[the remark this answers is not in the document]
vistaarjuneja: done

command/harness/destroy.go
```diff
@@ -20,7 +27,29 @@ func HandleDestroy(ctx context.Context, r *VMCleanupRequest, s store.StageOwnerS
 	if r.StageRuntimeID == "" {
 		return ierrors.NewBadRequestError("mandatory field 'stage_runtime_id' in the request body is empty")
 	}
+	// We do retries on destroy in case a destroy call comes while an initialize call is still happening.
+	cnt := 0
+	b := createBackoff(destroyTimeout)
```
shubham149, line 32: What's the max retries that this loop will carry out in case vm is never found?

vistaarjuneja:
```func createBackoff(maxElapsedTime time.Duration) *backoff.ExponentialBackOff {
	exp := backoff.NewExponentialBackOff()
	exp.MaxElapsedTime = maxElapsedTime
	return exp
}

PR #257, opened 2023-03-16
Add prefix to hostname while setting default runner name

rutvijmehta-harness:
If the hostname starts with a number (example `0b68be1f3feb`), GCP API call fails with the below error
`googleapi: Error 400: Invalid value for field 'resource.name': '0b68be1f3feb-ubuntu-gcp-pgfir7t8-t0n4f'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)'`

# Commit Checklist

## The Basics

### DIFF
--- a/internal/drivers/nomad/driver.go
+++ b/internal/drivers/nomad/driver.go
@@ -136,9 +136,12 @@ func (p *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (*t
 		return nil, fmt.Errorf("scheduler: could not register job, err: %w", err)
 	}
 	// If resources don't become available in `resourceJobTimeout`, we fail the step
-	_, err = p.pollForJob(ctx, resourceJobID, logr, resourceJobTimeout, true, []JobStatus{Running, Dead})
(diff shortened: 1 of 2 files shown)


command/config/config.go
```diff
@@ -390,7 +390,8 @@ func FromEnviron() (EnvConfig, error) {
 		config.Runner.Environ = map[string]string{}
 	}
 	if config.Runner.Name == "" {
-		config.Runner.Name, _ = os.Hostname()
+		hostname, _ := os.Hostname()
+		config.Runner.Name = fmt.Sprintf("runner-%s", hostname)
```
shubham149, line 394: Add this only if host name is empty string or starts with non lower case alphabet

rutvijmehta-harness: Done

rutvijmehta-harness: Should we do `strings.ToLower(hostname)` by default since Google API has a hard check on that?

Regex in the error: `(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)`

shubham149:
Rebase the pr. It will fix the lint errors

PR #276, opened 2023-04-20
(feat) ability to support groupID in Anka build cloud. 

eoinmcafee00:


command/config/config.go
```diff
@@ -456,5 +449,42 @@ func (s *Instance) UnmarshalJSON(data []byte) error {
...
+	}
+
 	return json.Unmarshal(obj.Spec, s.Spec)
 }
+
+type AliasAnkaBuild AnkaBuild
+
+// UnmarshalJSON required for backwards compatibility
+// with the old registry_url field.
+func (m *AnkaBuild) UnmarshalJSON(data []byte) error {
```
tphoney, line 478: 2 functions with the same name is a little confusing.

command/config/config.go
```diff
@@ -424,16 +426,7 @@ func FromEnviron() (EnvConfig, error) {
 }
 
 // UnmarshalJSON implement the json.Unmarshaler interface.
```
tphoney, line 428: fix the comment

tphoney (changes requested):

command/config/config.go
eoinmcafee00: they need to be the same name because they are implementing the json.Unmarshaler interface for their respective types. When the JSON unmarshaling process encounters these types, it will use their custom UnmarshalJSON functions.

PR #284, opened 2023-04-25
handle logging, cleanup on bare metal

harshjain12:
Added logging improvement, purge handling as job logs are vanishing after failure and force cleanup.

# Commit Checklist

## The Basics

### DIFF
--- a/internal/drivers/nomad/driver.go
+++ b/internal/drivers/nomad/driver.go
@@ -468,7 +468,7 @@ func (p *config) destroyJob(vm, nodeID string) (job *api.Job, id string) {
 						Driver:    "raw_exec",
 						Config: map[string]interface{}{
 							"command": "/usr/bin/su",
-							"args":    []string{"-c", fmt.Sprintf("%s stop -f %s && %s rm -f %s", ignitePath, vm, ignitePath, vm)},
+							"args":    []string{"-c", fmt.Sprintf("%s stop %s && %s rm %s && %s stop -f %s && %s rm -f %s", ignitePath, vm, ignitePath, vm, ignitePath, vm, ignitePath, vm)},
 						},
 					},
 				},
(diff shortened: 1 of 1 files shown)


internal/drivers/nomad/driver.go
```diff
@@ -468,7 +468,7 @@ func (p *config) destroyJob(vm, nodeID string) (job *api.Job, id string) {
 						Driver:    "raw_exec",
 						Config: map[string]interface{}{
 							"command": "/usr/bin/su",
-							"args":    []string{"-c", fmt.Sprintf("%s stop %s && %s rm %s", ignitePath, vm, ignitePath, vm)},
+							"args":    []string{"-c", fmt.Sprintf("%s stop -f %s && %s rm -f %s", ignitePath, vm, ignitePath, vm)},
```
hemanthmantri, line 471: From the documentation, force `rm` will also `kill`(alias for `stop -f`) the VM. So we just need `ignite rm -f vm`

hemanthmantri: > To remove VMs in Ignite, use the following command:

># ignite rm my-vm
>The VM needs to not be running for this to succeed. Using the --force flag (-f for short) a running VM can also be removed, it will be killed before removal. This force option is also able to stop and remove zombie VMs.

> # ignite rm -f my-vm

hemanthmantri: https://github.com/weaveworks/ignite/blob/main/docs/usage.md#removing-a-vm

internal/drivers/nomad/driver.go
```diff
@@ -196,7 +196,7 @@ func (p *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (*t
 	err = p.checkTaskGroupStatus(initJobID, initTaskGroup)
 	if err != nil {
 		defer p.Destroy(context.Background(), []*types.Instance{instance}) //nolint:errcheck
-		return nil, fmt.Errorf("scheduler: init job failed with error: %s", err)
+		return nil, fmt.Errorf("scheduler: init job failed with error: %s on Ip: %s", err, ip)
```
hemanthmantri, line 199: s/Ip/IP

Is it okay to print IPs in the logs? We removed them from init log I think.

internal/drivers/nomad/driver.go
harshjain12: We need stop first for regular cleanup.

internal/drivers/nomad/driver.go
harshjain12: IP is already logged in logs. Currently it is hard to debug without ip details

internal/drivers/nomad/driver.go
hemanthmantri: Where is the regular cleanup happening? I only see force cleanup.

hemanthmantri (approved):

vistaarjuneja (approved):

PR #293, opened 2023-05-05
use certs cache instead of regenerating cert each time

vistaarjuneja:
# Commit Checklist

## The Basics

### DIFF
--- a/internal/drivers/manager.go
+++ b/internal/drivers/manager.go
@@ -25,6 +25,7 @@ type (
 	Manager struct {
 		globalCtx            context.Context
 		poolMap              map[string]*poolEntry
+		certsCache           *certs.CertsCache
 		strategy             Strategy
 		cleanupTimer         *time.Ticker
 		runnerName           string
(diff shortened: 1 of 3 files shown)


internal/drivers/manager.go
```diff
@@ -514,7 +516,7 @@ func (m *Manager) setupInstance(ctx context.Context, pool *poolEntry, inuse bool
 	var inst *types.Instance
 
 	// generate certs
-	createOptions, err := certs.Generate(m.runnerName)
+	createOptions, err := m.certsCache.Get(m.runnerName)
```
shubham149, line 519: With this, it is possible that customer1 knows the certs for any build in 2 hr duration. Can the customer make rest calls to a different machine (of other account) using these certs?

shubham149: If we can ensure that east west traffic in the cloud vendor is not allowed, then we can have such restriction.

vistaarjuneja: nice catch @shubham149, i missed that. let me check

PR #294, opened 2023-05-09
Add prefix to hostname while setting default runner name

rutvijmehta-harness:
If the hostname starts with a number (example `0b68be1f3feb`), GCP API call fails with the below error
`googleapi: Error 400: Invalid value for field 'resource.name': '0b68be1f3feb-ubuntu-gcp-pgfir7t8-t0n4f'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)'`

# Commit Checklist

## The Basics

### DIFF
--- a/command/config/config.go
+++ b/command/config/config.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strings"
 	"unicode"
+	utf8 "unicode/utf8"
 
 	"github.com/drone-runners/drone-runner-aws/types"
 
@@ -400,7 +401,8 @@ func FromEnviron() (EnvConfig, error) {
 	if config.Runner.Name == "" {
 		hostname, _ := os.Hostname()
 		hostname = strings.ToLower(hostname)
-		if hostname == "" || !unicode.IsLower([]rune(hostname)[0]) {
+		r, _ := utf8.DecodeRuneInString(hostname)
+		if hostname == "" || !unicode.IsLower(r) {
 			config.Runner.Name = fmt.Sprintf("runner-%s", hostname)
 		} else {
 			config.Runner.Name = hostname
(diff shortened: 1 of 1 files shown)


command/config/config.go
```diff
@@ -396,7 +399,14 @@ func FromEnviron() (EnvConfig, error) {
 		config.Runner.Environ = map[string]string{}
 	}
 	if config.Runner.Name == "" {
-		config.Runner.Name, _ = os.Hostname()
+		hostname, _ := os.Hostname()
+		hostname = strings.ToLower(hostname)
+		r, _ := utf8.DecodeRuneInString(hostname)
```
shubham149, line 404:
```
# check if first character in hostname is lowercase
r, size := utf8.DecodeRuneInString(hostname)
if hostname == "" || !(size > 0 && unicode.IsLower(r) ) {
  config.Runner.Name = fmt.Sprintf("runner-%s", hostname)
}
```

rutvijmehta-harness: `utf8.DecodeRuneInString` returns `65533`(which is not lower case) if `hostname` is an empty string. The above code won't work if any other character in the hostname is upper string since GCP regex is `'(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)'`

command/config/config.go
```diff
@@ -396,7 +399,14 @@ func FromEnviron() (EnvConfig, error) {
 		config.Runner.Environ = map[string]string{}
 	}
 	if config.Runner.Name == "" {
-		config.Runner.Name, _ = os.Hostname()
+		hostname, _ := os.Hostname()
+		hostname = strings.ToLower(hostname)
```
shubham149, line 403: Add a comment after this line:
```
# check if first character in hostname is lowercase
```

rutvijmehta-harness: Done

shubham149 (approved):

PR #300, opened 2023-05-16
(maint) clean up us-used vms on maintenance mode

eoinmcafee00:


command/harness/dlite/handler.go
```diff
@@ -36,11 +37,15 @@ func Handler(p *poller.Poller) http.Handler {
 	return r
 }
 
-func handleEnable(p *poller.Poller) http.HandlerFunc {
+func handleEnable(p *poller.Poller, manager *drivers.Manager) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		p.SetFilter(func(ev *client.TaskEvent) bool {
 			return ev.TaskType != initTask
 		})
+		err := manager.CleanPools(r.Context(), false, true)
```
vistaarjuneja, line 45: It seems this function does not use a lock anywhere. It's important to lock before doing a list of VMs since after that a VM could become inuse and we might delete a inuse VM. Can we add locking in `CleanPools`?

eoinmcafee00: done

vistaarjuneja (approved):

PR #311, opened 2023-06-23
[CI-8232]: added logic for not cleaning vms in case setup fails

devkimittal:
…p (#310)

* [CI-8549]: Fix bazel test rule logic for fallback, engine version bump

* [CI-8549]: Fix bazel test rule logic for fallback, engine version bump

# Commit Checklist

## The Basics

### DIFF
--- a/command/debug/debug_state.go
+++ b/command/debug/debug_state.go
@@ -1,0 +1,54 @@
+package debug
+
+import (
+	"sync"
+)
+
+var (
+	state *State
+	once  sync.Once
+)
+
+// DebugState stores the variables map required for debugging vm flow.
+type State struct {
+	mu       sync.Mutex
+	retainVM map[string]bool
+}
+
+func (s *State) Get(stageRuntimeID string) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	val, ok := s.retainVM[stageRuntimeID]
+	if ok {
+		return val
+	}
+	return false
+}
+
+func (s *State) Set(stageRuntimeID string, retainVM bool) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if _, ok := s.retainVM[stageRuntimeID]; !ok {
+		s.retainVM[stageRuntimeID] = true
+	}
+	s.retainVM[stageRuntimeID] = retainVM
+}
+
+func (s *State) Delete(stageRuntimeID string) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	delete(s.retainVM, stageRuntimeID)
+}
+
+func GetState() *State {
+	once.Do(func() {
+		state = &State{
+			mu:       sync.Mutex{},
+			retainVM: make(map[string]bool),
+		}
+	})
+	return state
+}
--- a/command/harness/setup.go
+++ b/command/harness/setup.go
@@ -10,6 +10,7 @@ import (
 	"github.com/drone-runners/drone-runner-aws/internal/oshelp"
 
 	"github.com/drone-runners/drone-runner-aws/command/config"
+	"github.com/drone-runners/drone-runner-aws/command/debug"
 	"github.com/drone-runners/drone-runner-aws/engine/resource"
 	"github.com/drone-runners/drone-runner-aws/internal/drivers"
 	"github.com/drone-runners/drone-runner-aws/internal/lehelper"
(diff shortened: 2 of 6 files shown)


command/harness/setup.go
```diff
@@ -214,6 +215,12 @@ func HandleSetup(ctx context.Context, r *SetupVMRequest, s store.StageOwnerStore
...
 
+	deleteInfoMap := make(map[string]string)
+	deleteInfoMap["shouldDelete"] = "false"
+	if env.Runner.CleanupDebug {
+		util.CreateEnvState().Add(stageRuntimeID, deleteInfoMap)
+	}
+
 	setupResponse, err := client.Setup(ctx, &r.SetupRequest)
 	if err != nil {
 		go cleanUpFn(true)
```
shubham149, line 226: This will destroy the vm.

shubham149:
where is the code present to not delete the vm in case setup fails.

command/harness/setup.go
devkimittal: this will ultimately go to manager.destroy. I have put the logic there to not destroy

command/harness/setup.go
```diff
@@ -222,5 +230,11 @@ func HandleSetup(ctx context.Context, r *SetupVMRequest, s store.StageOwnerStore
 
 	logr.WithField("response", fmt.Sprintf("%+v", setupResponse)).Traceln("VM setup is complete")
 
+	deleteInfoMap["shouldDelete"] = "true"
+	if env.Runner.CleanupDebug {
+		logr.Traceln("shouldDelete set to true")
+		util.CreateEnvState().Add(stageRuntimeID, deleteInfoMap)
```
shubham149, line 236: envState is used for step output envs. We should not use the same for this purpose.

command/harness/setup.go
```diff
@@ -222,5 +230,11 @@ func HandleSetup(ctx context.Context, r *SetupVMRequest, s store.StageOwnerStore
 
 	logr.WithField("response", fmt.Sprintf("%+v", setupResponse)).Traceln("VM setup is complete")
 
+	deleteInfoMap["shouldDelete"] = "true"
```
shubham149, line 233: we should use a bool instead of string for true or false.

internal/drivers/manager.go
```diff
@@ -355,7 +356,12 @@ func (m *Manager) Destroy(ctx context.Context, poolName, instanceID string) erro
 	if err != nil {
 		return err
 	}
-
+	stageRuntimeID := instance.Stage
+	deleteInfoMap := util.CreateEnvState().Get(stageRuntimeID)
+	if deleteInfoMap["shouldDelete"] == "false" {
```
shubham149, line 361: what happens if shouldDelete is not present in the map

devkimittal: nothing, the older flow. Vm will be deleted.

command/harness/setup.go
devkimittal: I don't see any issue if we reuse this here. Besided it's a just a custom tag for customer test

command/harness/setup.go
devkimittal: create env state takes a string to string mapping map

command/harness/setup.go
shubham149: every env output for a step will have these values which should not be present.

devkimittal: got it, Will create a new class for it.

command/debug/debug_state.go
```diff
@@ -0,0 +1,56 @@
...
+	defer s.mu.Unlock()
+
+	delete(s.env, stageRuntimeID)
+}
+
+func GetDebugState() *DebugState {
+	once.Do(func() {
+		state = &DebugState{
+			mu:  sync.Mutex{},
+			env: make(map[string]map[string]string),
```
devkimittal, line 52: @shubham149 create a new class for storing this info. Im making the map as string->string. As later we might need to store other info as well.

shubham149: we should not create it thinking of future. Current entry would be better having just a bool. If required later, it can be transformed to a different structure.

cleanupVm: make(map[string]bool)

internal/drivers/manager.go
```diff
@@ -356,6 +357,13 @@ func (m *Manager) Destroy(ctx context.Context, poolName, instanceID string) erro
 		return err
 	}
 
+	stageRuntimeID := instance.Stage
+	cleanupVm := debugState.GetDebugState().Get(stageRuntimeID)
```
shubham149, line 361: check if entry is present or use field name as retainVm so that its default value means destroying the vm

devkimittal: value is coming as true if the entry is not present.

shubham149: Better to use retain vm instead since we mostly assume that default value is false.

devkimittal: makes sense, making the change

vistaarjuneja:
could we not store this in the instance DB itself? whether it needs to be destroyed or not @devkimittal @shubham149

PR #321, opened 2023-07-20
fix: clean other pools if one fails

devkimittal:
# Commit Checklist

## The Basics

### DIFF
--- a/internal/drivers/manager.go
+++ b/internal/drivers/manager.go
@@ -411,14 +411,16 @@ func (m *Manager) cleanPool(ctx context.Context, pool *poolEntry, destroyBusy, d
 }
 
 func (m *Manager) CleanPools(ctx context.Context, destroyBusy, destroyFree bool) error {
+	var returnError error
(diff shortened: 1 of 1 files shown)


internal/drivers/manager.go
```diff
@@ -414,7 +414,7 @@ func (m *Manager) CleanPools(ctx context.Context, destroyBusy, destroyFree bool)
 	for _, pool := range m.poolMap {
 		err := m.cleanPool(ctx, pool, destroyBusy, destroyFree)
 		if err != nil {
-			return err
+			logrus.Errorf("failed to clean pool %s with error: %s", pool, err)
```
vistaarjuneja, line 417: probably still return the error on L421 instead of returning nil. But continue to try to clean all the pools

shubham149, line 417: Return error on L421 if any one of the pool failed to cleanup

devkimittal: done

vistaarjuneja (approved):

PR #340, opened 2023-08-21
(bug) check health end point of service for mac before returning success

eoinmcafee00:


vistaarjuneja:
@eoinmcafee00 one thing I am concerned about with this change is the behaviour for setup now.
We have seen instances before where retry health check hangs. What would happen with those cases now that we skip health check? Would the setup API call hang?
Also if setup is going to hang anyway maybe it's better to have a health check as we would not be able to proceed in that case? So maybe we add health check in both places

PR #344, opened 2023-08-25
ci-9308 - improve provision logic, run health check on allocation 

eoinmcafee00:


tphoney (commented):
mockClientFactory feels like its over engineered, and isnt needed as an interface. But that is just my feeling. It may be the only way to allow this kind of testing. 
 I would recommend someone else looks at this, with more experience in complex mock testing

internal/drivers/manager.go
```diff
@@ -337,15 +341,57 @@ func (m *Manager) Provision(ctx context.Context, poolName, serverName, ownerID s
 	}
 	pool.Unlock()
 
+	// do health check
+	if healthCheckErr := m.performHealthCheck(ctx, inst, env); healthCheckErr != nil {
```
vistaarjuneja, line 345: Isn't this doing a health check on the VM already in the pool? And then there's another health check happening in `setup.go`? That doesn't seem right unless I'm missing something.
We should do health check before adding to the pool, and health check should only be happening once

eoinmcafee00: I had removed the health check from setup.go but @tphoney requested it stay, so I put it back

PR #346, opened 2023-08-29
[CI-7513]: Abort any running steps when cleanup request is received

raghavharness:
Added logic to abort and running steps when the cleanup request comes for a stage.

### DIFF
--- a/command/harness/dlite/cleanup.go
+++ b/command/harness/dlite/cleanup.go
@@ -40,6 +40,7 @@ func (t *VMCleanupTask) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		httphelper.WriteBadRequest(w, err)
 		return
 	}
+	ctxState().Delete(req.StageRuntimeID)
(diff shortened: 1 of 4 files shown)


vistaarjuneja:
@raghavharness nice work! LGTM.
One thing to think about (and we should create a ticket for this): With the new design we will remove the stickiness of pipelines. So execute and cleanup tasks will be able to be routed to a different dlite than the setup. We have to come up with a good way to handle this in the newer model. Maybe have the retry poll step check the database instance state after a few retries to see if the instance has disappeared or not. Let's create a ticket for this and link it to the scalability epic so that we do not miss it.

vistaarjuneja (approved):

raghavharness:
> @raghavharness nice work! LGTM. One thing to think about (and we should create a ticket for this): With the new design we will remove the stickiness of pipelines. So execute and cleanup tasks will be able to be routed to a different dlite than the setup. We have to come up with a good way to handle this in the newer model. Maybe have the retry poll step check the database instance state after a few retries to see if the instance has disappeared or not. Let's create a ticket for this and link it to the scalability epic so that we do not miss it.

Created ticket for this - https://harness.atlassian.net/browse/CI-9439

PR #350, opened 2023-09-08
add recover middleware and some enhances

raghavharness:
Added recovery middleware for poller and remove nil pointer exception if hibernate start operation fails. If nil pointer comes, dlite restarts.

### DIFF
--- a/command/harness/dlite/handler.go
+++ b/command/harness/dlite/handler.go
@@ -1,6 +1,7 @@
 package dlite
 
 import (
+	"fmt"
 	"io"
 	"net/http"
 
@@ -11,6 +12,7 @@ import (
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
 	"github.com/wings-software/dlite/client"
+	"github.com/wings-software/dlite/httphelper"
 	"github.com/wings-software/dlite/poller"
 )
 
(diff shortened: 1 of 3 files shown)


command/harness/dlite/handler.go
```diff
@@ -70,3 +72,21 @@ func handleStatus(p *poller.Poller) http.HandlerFunc {
 		io.WriteString(w, enabledStatus) //nolint: errcheck
 	}
 }
+
+func pollerMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer func() {
+			if r := recover(); r != nil {
+				// Handle the panic here, log it, or respond with an error message.
+				fmt.Println("Panic occurred:", r)
```
vistaarjuneja, line 81: Nit: Use logger instead of fmt

command/harness/dlite/handler.go
```diff
@@ -70,3 +72,21 @@ func handleStatus(p *poller.Poller) http.HandlerFunc {
...
+				errMsg := "Unable to handle request, something went wrong."
```
vistaarjuneja, line 82: you could do err = `fmt.Errorf("http: panic: %v", r)`

command/harness/setup.go
```diff
@@ -219,24 +219,28 @@ func handleSetup(
...
+			out, logErr := poolManager.InstanceLogs(context.Background(), pool, instanceID)
 			if logErr != nil {
 				logr.WithError(logErr).Errorln("failed to fetch console output logs")
 			} else {
 				// Serial console output is limited to 60000 characters since stackdriver only supports 64KB per log entry
 				l := math.Min(float64(len(out)), 60000) //nolint:gomnd
-				logrus.WithField("id", instance.ID).
-					WithField("instance_name", instance.Name).
-					WithField("ip", instance.Address).
+				logrus.WithField("id", instanceID).
```
harshjain12, line 235: can we add account id ?

harshjain12 (approved):

command/harness/setup.go
raghavharness: Will change from logrus to logr in next pr so that it takes values from context

PR #354, opened 2023-09-13
[CI-9439]: Changes to remove dlite stickyness

raghavharness:
Changes to remove dlite stickyness


command/harness/dlite/dlite.go
```diff
@@ -103,29 +104,31 @@ func (c *dliteCommand) run(*kingpin.ParseContext) error {
...
-	poolConfig, err := harness.SetupPool(ctx, &c.env, c.poolManager, c.poolFile)
-	defer harness.Cleanup(&c.env, c.poolManager) //nolint: errcheck
+	poolConfig, err := c.setupPool(ctx)
+	defer harness.Cleanup(&c.env, c.poolManager, true, true) //nolint: errcheck
 	if err != nil {
-		logrus.WithError(err).Error("could not setup pool")
 		return err
 	}
 
+	if env.Database.Postgres.Enabled {
```
vistaarjuneja, line 116: I feel like now is the right time to move to `wire` in this project and do dependency injection. Maybe we can do it subsequently but all this would be so much easier to manage through wire.

command/harness/dlite/dlite.go
```diff
@@ -103,29 +104,31 @@ func (c *dliteCommand) run(*kingpin.ParseContext) error {
...
-
-	c.stageOwnerStore = stageOwnerStore
-	c.poolManager = drivers.New(ctx, instanceStore, &c.env)
+	// Initialize metrics
+	c.registerMetrics()
 
-	poolConfig, err := harness.SetupPool(ctx, &c.env, c.poolManager, c.poolFile)
-	defer harness.Cleanup(&c.env, c.poolManager) //nolint: errcheck
+	poolConfig, err := c.setupPool(ctx)
+	defer harness.Cleanup(&c.env, c.poolManager, true, true) //nolint: errcheck
```
vistaarjuneja, line 111: Do we need to call defer cleanup on the pool manager as well? Only one or the other would be used right?

command/harness/dlite/dlite.go
```diff
@@ -137,14 +140,21 @@ func (c *dliteCommand) run(*kingpin.ParseContext) error {
 
 	g.Go(func() error {
 		<-ctx.Done()
-		return harness.Cleanup(&c.env, c.poolManager)
+		err = harness.Cleanup(&c.env, c.poolManager, true, true)
+		if c.env.Database.Postgres.Enabled {
```
vistaarjuneja, line 144: +1, do we need to call cleanup for both pool manager and distributed pool manager?
Also, this naming is confusing -> if postgres is enabled, we are using "distributed mode". Can we instead keep all the database structs the same (reuse) and then we can have a separate flag for a mode which assumes that the database is being shared across multiple instances. `<>_SHARE_DB_INSTANCE`
"Postgres" on its own doesn't indicate database sharing.

metric/builds.go
```diff
@@ -59,57 +73,68 @@ func FailedBuildCount() *prometheus.CounterVec {
 			Name: "harness_ci_pipeline_execution_errors_total",
 			Help: "Total number of pipeline executions which failed due to system errors",
 		},
-		[]string{"pool_id", "os", "arch", "driver"},
+		[]string{"pool_id", "os", "arch", "driver", "distributed"},
```
vistaarjuneja, line 76: why do we want to store `distributed` in the metric? How is it useful for the build count?

metric/builds.go
```diff
@@ -41,14 +51,18 @@ func ConvertBool(b bool) string {
...
+
 // BuildCount provides metrics for total number of pipeline executions (failed + successful)
 func BuildCount() *prometheus.CounterVec {
 	return prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Name: "harness_ci_pipeline_execution_total",
 			Help: "Total number of completed pipeline executions (failed + successful)",
 		},
-		[]string{"pool_id", "os", "arch", "driver"},
+		[]string{"pool_id", "os", "arch", "driver", "distributed"},
```
vistaarjuneja, line 65: +1, why do we need distributed?

metric/builds.go
```diff
@@ -59,57 +73,68 @@ func FailedBuildCount() *prometheus.CounterVec {
...
 // We are just using two labels at the moment which should be pretty small
-func RunningPerAccountCount(instanceStore store.InstanceStore) *prometheus.GaugeVec {
+func RunningPerAccountCount() *prometheus.GaugeVec {
 	return prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "harness_ci_pipeline_per_account_running_executions",
 			Help: "Total number of running executions per account",
 		},
-		[]string{"owner_id", "os"},
+		[]string{"owner_id", "os", "distributed"},
```
vistaarjuneja, line 100: +1 (applies to all metrics in this file) why do we need distributed? Only add fields to metrics which are absolutely needed and make sense to query on.
With prometheus every new entry you store here multiplies the size of the timeseries database

metric/builds.go
```diff
@@ -59,57 +73,68 @@ func FailedBuildCount() *prometheus.CounterVec {
...
-		for _, i := range instances {
-			l := label{os: i.OS, arch: i.Arch, state: string(i.State), poolID: i.Pool, driver: string(i.Provider)}
-			if i.OwnerID != "" {
-				runningPerAccountMetric.WithLabelValues(i.OwnerID, i.OS).Inc()
+func (m *Metrics) UpdateRunningCount(ctx context.Context) {
+	go func() {
+		for {
+			time.Sleep(dbInterval)
+			wg := &sync.WaitGroup{}
+			for _, ms := range m.stores {
```
vistaarjuneja, line 109: why do we need multiple stores here and a wait group?
For single mode, it would iterate over sqlite and for distributed mode over postgres - only difference would be the way it's querying the running builds right?

metric/builds.go
```diff
@@ -59,57 +73,68 @@ func FailedBuildCount() *prometheus.CounterVec {
...
+				go m.updateRunningCount(ctx, ms, wg)
 			}
-			m[l]++
+			wg.Wait()
+			m.RunningPerAccountCount.Reset()
```
vistaarjuneja, line 113: I think this will cause issues.
Reset will set all the cardinalities to 0. So imagine you had a mac build and prometheus did not fetch the metrics yet. In that case, in the next iteration you would reset it to zero and its existence will be gone.
Instead, prometheus by itself will remove entries which have been stale for more than 5 minutes. Read this: https://prometheus.io/docs/prometheus/latest/querying/basics/#staleness

store/database/migrate/postgres/0001_create_table_instances.up.sql
```diff
@@ -11,7 +11,7 @@ CREATE TABLE IF NOT EXISTS instances (
     ,instance_size      VARCHAR(50)
     ,instance_platform  VARCHAR(50)
     ,instance_arch      VARCHAR(50)
-    ,instance_stage     INTEGER
+    ,instance_stage     VARCHAR(50)
```
vistaarjuneja, line 14: We should not be changing existing schemas

command/harness/dlite/dlite.go
```diff
@@ -24,19 +24,20 @@ import (
...
+	poolFile               string
+	poolManager            drivers.IManager
+	distributedPoolManager drivers.IManager
+	metrics                *metric.Metrics
 }
 
 func RegisterDlite(app *kingpin.Application) {
 	c := new(dliteCommand)
 
 	c.poolManager = &drivers.Manager{}
```
vistaarjuneja, line 39: Can we set the pool manager just once? That would avoid having `poolManager` and `distributedPoolManager` everywhere even though only one will ever be used. If you need to check which case it is, you could do so from the env (`if SHARED_POOL...`)

command/harness/dlite/dlite.go
```diff
@@ -171,3 +181,52 @@ func (c *dliteCommand) run(*kingpin.ParseContext) error {
...
+		Store: instanceStore,
+		Query: &types.QueryParams{
+			RunnerName: c.env.Runner.Name,
+		},
+		Distributed: true,
+	})
+	return poolConfig, nil
+}
+
+func (c *dliteCommand) getPoolManager(distributed bool) drivers.IManager {
```
vistaarjuneja, line 227: Can we just use a single pool manager `c.poolManager` and set it up once in the beginning. This would also help once we do dependency injection and is a cleaner approach.

internal/drivers/distributed_manager.go
```diff
@@ -0,0 +1,170 @@
...
+			"distributed dlite: max age of used instances (set to %.2fmin) should be less than max age of free instances (set to %.2fmin)",
+			maxAgeBusy.Minutes(), maxAgeFree.Minutes())
+	}
+
+	if d.cleanupTimer != nil {
+		panic("distributed dlite: purger already started")
+	}
+
+	t := time.Duration(maxAgeBusy.Minutes() * 0.9 * float64(time.Minute))
+	d.cleanupTimer = time.NewTicker(t)
```
vistaarjuneja, line 83: Add `defer d.cleanupTimer.Stop()`. If it's not there in the existing manager, can we add it there as well?

command/harness/destroy.go
```diff
@@ -45,22 +46,28 @@ func HandleDestroy(ctx context.Context, r *VMCleanupRequest, s store.StageOwnerS
...
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(duration):
+			_, err := handleDestroy(ctx, r, s, env, poolManager, metrics, cnt, logr)
+			if err != nil {
+				logr.WithError(err).Errorln("could not destroy VM")
+				if duration == backoff.Stop {
+					return err
+				}
+				time.Sleep(duration)
```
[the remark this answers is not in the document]
raghavharness: Actually I didn't change it. It was already like this.

internal/drivers/distributed_manager.go
raghavharness: But wouldn't it stop the timer ?

command/harness/destroy.go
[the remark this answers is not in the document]
raghavharness: But I can remove since this looks unnecessary

command/harness/destroy.go
```diff
@@ -45,22 +46,28 @@ func HandleDestroy(ctx context.Context, r *VMCleanupRequest, s store.StageOwnerS
...
-		_, err := handleDestroy(ctx, r, s, env, poolManager, metrics, cnt, logr)
-		if err != nil {
-			logr.WithError(err).Errorln("could not destroy VM")
-			if duration == backoff.Stop {
-				return err
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(duration):
```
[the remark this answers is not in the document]
raghavharness: Thanks for sharing this, it's a nice read.

internal/drivers/distributed_manager.go
vistaarjuneja: you are right, maybe move it inside the go func and defer stop there

raghavharness: But even in the go routine if we stop the next tick won't happen no ? The purger ticker should remain running.

command/config/config.go
```diff
@@ -363,6 +363,11 @@ type EnvConfig struct {
 	Database struct {
 		Driver     string `envconfig:"DRONE_DATABASE_DRIVER" default:"sqlite3"`
 		Datasource string `envconfig:"DRONE_DATABASE_DATASOURCE" default:"database.sqlite3"`
+		Postgres   struct {
```
vistaarjuneja, line 366: Can we move this struct out to `Distributed`? Maybe something like
```
   DistributedMode {
        Enabled
        Driver
        Datasource
   }
```

internal/drivers/distributed_manager.go
raghavharness: Ohh I see where you mean

store/database/migrate/postgres/0001_create_table_instances.up.sql
vistaarjuneja: @raghavharness are we ok with this?

command/config/config.go
raghavharness: Sure

raghavharness: Done

internal/drivers/distributed_manager.go
raghavharness: Done

command/harness/destroy.go
```diff
@@ -45,22 +46,32 @@ func HandleDestroy(ctx context.Context, r *VMCleanupRequest, s store.StageOwnerS
 	b := createBackoff(destroyTimeout)
 	for {
 		duration := b.NextBackOff()
-		_, err := handleDestroy(ctx, r, s, env, poolManager, metrics, cnt, logr)
-		if err != nil {
-			logr.WithError(err).Errorln("could not destroy VM")
-			if duration == backoff.Stop {
-				return err
+		timer := time.NewTimer(duration)
```
vistaarjuneja, line 49: Move this timer out of the for, we shouldn't create a new timer each time.
```
timer := time.NewTimer(duration)
for {
       if !timer.Stop() {
              <-timer.C
       }
       timer.Reset(duration)
       select {
             case <-ctx.Done():
                    break / return
             case <-timer.C:
                    ///
       }
}
```

vistaarjuneja (approved):

PR #358, opened 2023-09-22
CI-9662: Added google dns for windows hosted

devkimittal:
# Commit Checklist

## The Basics

### DIFF
--- a/command/harness/setup.go
+++ b/command/harness/setup.go
@@ -43,7 +43,7 @@ type SetupVMResponse struct {
 }
 
 var (
-	healthCheckTimeout = 3 * time.Minute
+	healthCheckTimeout = 5 * time.Minute
 	freeAccount        = "free"
 )
 
--- a/internal/cloudinit/cloudinit.go
+++ b/internal/cloudinit/cloudinit.go
@@ -29,6 +29,7 @@ type Params struct {
 	HarnessTestBinaryURI string
 	PluginBinaryURI      string
 	Tmate                types.Tmate
+	IsHosted             bool
 }
 
 var funcs = map[string]interface{}{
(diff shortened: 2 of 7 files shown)


internal/cloudinit/cloudinit.go
```diff
@@ -484,6 +484,10 @@ Invoke-WebRequest -Uri "{{ .LiteEnginePath }}/lite-engine-{{ .Platform.OS }}-{{
 New-NetFirewallRule -DisplayName "ALLOW TCP PORT 9079" -Direction inbound -Profile Any -Action Allow -LocalPort 9079 -Protocol TCP
 Start-Process -FilePath "C:\Program Files\lite-engine\lite-engine.exe" -ArgumentList "server --env-file=` + "`" + `"C:\Program Files\lite-engine\.env` + "`" + `"" -RedirectStandardOutput "{{ .LiteEngineLogsPath }}" -RedirectStandardError "C:\Program Files\lite-engine\log.err"
 
+if ($env:COMPUTERNAME -eq "PACKER-63BBB9DE") {
```
devkimittal, line 487: we just want to add this for our hosted vms. Since we don't pass any flag while we create the vm, im using COMPUTERNAME env variable to recognise it. This is same across all the windows hosted vms.

raghavharness (approved):

command/harness/setup.go
```diff
@@ -42,7 +42,7 @@ type SetupVMResponse struct {
 }
 
 var (
-	healthCheckTimeout = 3 * time.Minute
+	healthCheckTimeout = 5 * time.Minute
```
devkimittal, line 45: @vistaarjuneja increased this because windows builds are taking more than 3 minutes in some cases. had few failures in local testing.

internal/drivers/util.go
```diff
@@ -0,0 +1,16 @@
+package drivers
+
+import (
+	"context"
+
+	"github.com/drone-runners/drone-runner-aws/types"
+)
+
+func isHosted(ctx context.Context) bool {
```
vistaarjuneja, line 9: move this to `harness/context` instead as it's only relevant to harness

types/types.go
```diff
@@ -7,6 +7,12 @@ import (
 type InstanceState string
 type DriverType string
 
+type ContextKey string
```
vistaarjuneja, line 10: Declare this in `harness/context` and keep the defs there

types/types.go
```diff
@@ -7,6 +7,12 @@ import (
...
+
+const (
+	Hosted = ContextKey("hosted")
```
vistaarjuneja, line 13: Also, just use empty struct like:
```
type hostedKey struct{}

context.WithValue(ctx, hostedKey, true)
```

internal/drivers/util.go
devkimittal: this is used by manager.go, putting it in harness/context will make it cyclic dependency.

types/types.go
devkimittal: this is created to reuse it in future.

vistaarjuneja (approved):

PR #382, opened 2023-11-03
fix: [CI-9889]: Refactored code to handle fallback cleanups

raghavharness:
Refactored code to handle fallback cleanups.


vistaarjuneja (approved):
looks good! just make sure the abort case is tested

PR #422, opened 2024-02-13
fix: [CI-11237]: modified retry commands

ShobhitSingh11:
# Commit Checklist

## The Basics

### DIFF
--- a/internal/cloudinit/cloudinit.go
+++ b/internal/cloudinit/cloudinit.go
@@ -97,7 +97,7 @@ swapon /swapfile
 echo "done setting up swap space"
 
 echo "downloading lite engine binary"
-/usr/bin/wget --retry-connrefused --retry-on-host-error --retry-on-http-error=503,404,429 --tries=10 --waitretry=10 "{{ .LiteEnginePath }}/lite-engine-{{ .Platform.OS }}-{{ .Platform.Arch }}" -O /usr/bin/lite-engine
+/usr/bin/wget --retry-connrefused --retry-on-host-error --retry-on-http-error=503,404,429 --tries=10 --waitretry=10 "{{ .LiteEnginePath }}/lite-engine-{{ .Platform.OS }}-{{ .Platform.Arch }}" -O /usr/bin/lite-engine || wget --retry-connrefused --tries=10 --waitretry=10 -nv --debug "{{ .LiteEnginePath }}/lite-engine-{{ .Platform.OS }}-{{ .Platform.Arch }}" -O /usr/bin/lite-engine
 echo "done downloading lite engine binary"
 chmod 777 /usr/bin/lite-engine
 touch $HOME/.env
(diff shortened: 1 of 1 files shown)


internal/cloudinit/cloudinit.go
```diff
@@ -97,20 +97,20 @@ swapon /swapfile
 echo "done setting up swap space"
 
 echo "downloading lite engine binary"
-/usr/bin/wget --retry-connrefused --retry-on-host-error --retry-on-http-error=503,404,429 --tries=10 --waitretry=10 "{{ .LiteEnginePath }}/lite-engine-{{ .Platform.OS }}-{{ .Platform.Arch }}" -O /usr/bin/lite-engine
+/usr/bin/wget --retry-connrefused --retry-on-host-error --retry-on-http-error=503,404,429 --tries=10 --waitretry=10 "{{ .LiteEnginePath }}/lite-engine-{{ .Platform.OS }}-{{ .Platform.Arch }}" -O /usr/bin/lite-engine || /usr/bin/wget --retry-connrefused --tries=10 --waitretry=10 -nv --debug "{{ .LiteEnginePath }}/lite-engine-{{ .Platform.OS }}-{{ .Platform.Arch }}" -O /usr/bin/lite-engine
```
devkimittal, line 100: let's refactor it a bit, put ` {{ .LiteEnginePath }}/lite-engine-{{ .Platform.OS }}-{{ .Platform.Arch }}" -O /usr/bin/lite-engine` in variable and then do `wget --retry-connrefused --retry-on-host-error --retry-on-http-error=503,404,429 --tries=10 --waitretry=10 var1`
Basically clean it up a little at all the places

ShobhitSingh11: The reason we are doing it this way because we are not doing only for LE there are plugins, splitTests and the path 
"/usr/bin/lite-engine" is different for some other LE binary downloads commands, same for plugin as well. So ultimately we will end up with many such var strings

devkimittal: We will have one such variable per wherever we are using this wget || thing and that's alright I guess, as long as it helps removing duplicate code.

internal/cloudinit/cloudinit.go
```diff
@@ -97,20 +103,20 @@ swapon /swapfile
 echo "done setting up swap space"
 
 echo "downloading lite engine binary"
-/usr/bin/wget --retry-connrefused --retry-on-host-error --retry-on-http-error=503,404,429 --tries=10 --waitretry=10 "{{ .LiteEnginePath }}/lite-engine-{{ .Platform.OS }}-{{ .Platform.Arch }}" -O /usr/bin/lite-engine
+/usr/bin/wget --retry-connrefused --retry-on-host-error --retry-on-http-error=503,404,429 --tries=10 --waitretry=10 ` + liteEngineUsrBinPath + ` || /usr/bin/wget --retry-connrefused --tries=10 --waitretry=10 -nv --debug ` + liteEngineUsrBinPath + `
```
devkimittal, line 106: do we need this '+' at the end?

ShobhitSingh11: Since we are adding one string in other format is like this ` + string +` else it will give compilation error

devkimittal (approved):

internal/cloudinit/cloudinit.go
ShobhitSingh11: Ref:

raghavharness (approved):

PR #424, opened 2024-02-21
add timeout for gcp zone operation get

raghavharness:
We saw some executions where the gcp zone operation get call is getting stuck for default timeout (10m). Have added smaller timeout on that api call so that it can retry again.

### DIFF
--- a/internal/drivers/google/driver.go
+++ b/internal/drivers/google/driver.go
@@ -545,10 +545,10 @@ func (p *config) waitZoneOperation(ctx context.Context, name, zone string) error
 			return ctx.Err()
 		default:
 		}
-		ctx_gcp, cancel := context.WithTimeout(ctx, operationGetTimeout*time.Second)
-		defer cancel()
+		ctxGcp, cancel := context.WithTimeout(ctx, operationGetTimeout*time.Second)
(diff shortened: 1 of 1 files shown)


vistaarjuneja (approved):

vistaarjuneja:
LGTM

internal/drivers/google/driver.go
```diff
@@ -539,8 +540,15 @@ func (p *config) findInstanceZone(ctx context.Context, instanceID string) (
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+		ctxGcp, cancel := context.WithTimeout(ctx, operationGetTimeout*time.Second)
 		client := p.service
-		op, err := client.ZoneOperations.Get(p.projectID, zone, name).Context(ctx).Do()
+		op, err := client.ZoneOperations.Get(p.projectID, zone, name).Context(ctxGcp).Do()
+		cancel()
```
vistaarjuneja, line 551: good catch on using `cancel` since it's in an infinite for loop

raghavharness: Yeah linter helped here :)

PR #428, opened 2024-03-27
feat:[CI-11620]:configure disk size using create vm api

HenAmar:
# Commit Checklist

## The Basics

### DIFF
--- a/command/config/config.go
+++ b/command/config/config.go
@@ -97,6 +97,7 @@ type (
 		Tag           string `json:"tag,omitempty" yaml:"tag"`
 		AuthToken     string `json:"auth_token,omitempty" yaml:"auth_token"`
 		GroupID       string `json:"group_id,omitempty" yaml:"group_id"`
+		Disk          string `json:"disk,omitempty" yaml:"disk"`
 	}
 
 	Nomad struct {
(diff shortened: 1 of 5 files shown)


command/config/config.go
```diff
@@ -97,6 +97,7 @@ type (
 		Tag           string `json:"tag,omitempty" yaml:"tag"`
 		AuthToken     string `json:"auth_token,omitempty" yaml:"auth_token"`
 		GroupID       string `json:"group_id,omitempty" yaml:"group_id"`
+		Disk          string `json:"disk,omitempty" yaml:"disk"`
```
devkimittal, line 100: we will pass it from pool yaml?

HenAmar: yes, and if we don't it will go with the default.

devkimittal: what is the size that we are planning to put in prod?

devkimittal: SInce we just create one vm in a machine, we can use max memory available. also what is the default size right now?

HenAmar: default for sonoma is 128 and ventura is 256. I plan to put both with 256.

devkimittal (approved):

PR #453, opened 2024-06-06
fix: [CI-11826]: Added support Azure VM spin up for custom image

ShobhitSingh11:
## Description

We are supporting spinning up of Azure VM with custom image via pool.yml, initially it was supported for marketplace image only. 

Cases Tested
1. Keep the yaml as same as old one (backward compatibility ensured). VM was spinned up. (Image used here was market price image)
2. Changed the yaml by removing the parameters like sku, version, publisher, offer (generally used for market price image) and introduced new parameter "id" which is the full path of custom image. 
3. Trying to fill the parameters like sku, version, publisher, offer with custom image params, it will not work as it is not published image. Those params we won't find in VM spinned up custom image (By looking json response of VM)

### DIFF
--- a/command/config/config.go
+++ b/command/config/config.go
@@ -144,6 +144,7 @@ type (
 		Zones             []string          `json:"zones,omitempty" yaml:"zones,omitempty"`
 		Tags              map[string]string `json:"tags,omitempty" yaml:"tags,omitempty"`
 		SecurityGroupName string            `json:"security_group_name,omitempty" yaml:"security_group_name,omitempty"`
+		SecurityType      string            `json:"security_type,omitempty" yaml:"security_type,omitempty"`
 	}
 
 	AzureAccount struct {
--- a/command/config/config.go
+++ b/command/config/config.go
@@ -160,6 +160,7 @@ type (
 		Version   string `json:"version,omitempty"  yaml:"version,omitempty"`
 		Username  string `json:"username,omitempty"  yaml:"username,omitempty"`
 		Password  string `json:"password,omitempty"  yaml:"password,omitempty"`
+		ID        string `json:"id,omitempty"  yaml:"id,omitempty"`
 	}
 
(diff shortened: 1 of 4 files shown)


devkimittal (dismissed):

VivekHarness:
trigger messagecheck

command/config/config.go
```diff
@@ -160,6 +161,7 @@ type (
 		Version   string `json:"version,omitempty"  yaml:"version,omitempty"`
 		Username  string `json:"username,omitempty"  yaml:"username,omitempty"`
 		Password  string `json:"password,omitempty"  yaml:"password,omitempty"`
+		ID        string `json:"id,omitempty"  yaml:"id,omitempty"`
```
raghavharness, line 164: do go fmt also

internal/drivers/azure/driver.go
```diff
@@ -30,7 +30,9 @@ type config struct {
 
 	securityGroupName string
 
-	rootDir string
+	rootDir      string
+	ID           string
```
raghavharness, line 34: keep it as `id`

raghavharness (changes requested):

command/config/config.go
ShobhitSingh11: same result

raghavharness (approved):

ShobhitSingh11:
trigger messagecheck

internal/drivers/azure/driver.go
```diff
@@ -211,6 +221,13 @@ func (c *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (in
 		},
 	}
 
+	if c.id != "" {
```
vistaarjuneja, line 224: This should check for security type right?
```if c.securityType != ""```

ShobhitSingh11: We are taking ID as primary

PR #476, opened 2024-07-17
feat: [CDE-119]: Add task handling to spawn and cleanup VM for CDE/gitspaces on bare metal

vikyathharekal:
1. Adding task handlers for DLITE_CDE_VM_INITIALIZE_TASK and DLITE_CDE_VM_CLEANUP_TASK tasks which will be used to provision and destroy VM for CDE in bare metal machines
2. CDE manager will submit these tasks to manager. The task poller here will obtain these tasks from manager and start handling them.
3. Additional parameters as compared to the CI init task would be the secret used in agent to communicate to cde-manager and gcloud access token to download the agent binary itself.
4. Added init script for gitspaces which downloads and starts gitspaces agent in addition to what is already being done in linux script (which includes downloading and starting lite engine)
5. We are asking nomad to give us additional random free ports for gitspace agent and ssh/web. This information is passed in the delegate task params and handled dynamically. These are used in the VM port mapping to expose it outside (8083 -> gitspaces agent, 8088 -> ssh, 8089 -> web). These ports are returned in the response as well. We would eventually consume it in CDE manager to connect to the agent and ssh/web port will be used to remotely access dev container
6. I've added checks (isGitspaceRequest) as much as possible everywhere so as to not disturb the current CI flow.

**Overall idea is to leverage the existing code as much as possible and also make sure we don't disturb the existing CI code much so as to not cause any regressions**

# Commit Checklist

## The Basics


command/harness/dlite/gitspaces_init.go
```diff
@@ -0,0 +1,83 @@
...
+	"time"
+
+	"github.com/drone-runners/drone-runner-aws/command/harness"
+	"github.com/sirupsen/logrus"
+	"github.com/wings-software/dlite/client"
+	"github.com/wings-software/dlite/httphelper"
+)
+
+type GitspacesVMInitTask struct {
+	c *dliteCommand
```
n00bitax, line 18: can we give a better name than c?

vikyathharekal: Yes, can do it. I borrowed it from the CI task

dhruv-harness:
> We are asking nomad to give us 2 additional random free ports for gitspace agent and ssh. These are used in the VM port mapping to expose it outside (8083 -> gitspaces agent, 8088 -> ssh). These 2 ports are returned in the response as well. We would eventually consume it in CDE manager to connect to the agent and ssh port will be used to remotely access dev container
@vikyathharekal The number of ports required will be 3. 8083 is for the agent. 8088 for ssh and 8089 for VS Code Web server. Can we change the number of ports required on runtime?

command/harness/dlite/gitspaces_cleanup.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/wings-software/dlite/client"
+	"github.com/wings-software/dlite/httphelper"
+)
+
+type GitspacesVMCleanupTask struct {
+	c *dliteCommand
+}
+
+func (t *GitspacesVMCleanupTask) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, cancel := context.WithCancel(r.Context()) // TODO: Get this from http Request
```
dhruv-harness, line 20: Wont the ctx from the context already be cancellable?

command/harness/dlite/gitspaces_cleanup.go
```diff
@@ -0,0 +1,67 @@
...
+	"strconv"
+
+	"github.com/drone-runners/drone-runner-aws/command/harness"
+	"github.com/sirupsen/logrus"
+	"github.com/wings-software/dlite/client"
+	"github.com/wings-software/dlite/httphelper"
+)
+
+type GitspacesVMCleanupTask struct {
+	c *dliteCommand
```
dhruv-harness, line 16: Can we make the name a bit more descriptive so that it is easier to understand in the method body?
such as `cmd` or `dlcmd`

command/harness/dlite/gitspaces_cleanup.go
```diff
@@ -0,0 +1,67 @@
...
+	resp := GitspacesVMTaskExecutionResponse{
+		VMTaskExecutionResponse: VMTaskExecutionResponse{
+			CommandExecutionStatus: Success,
+			DelegateMetaInfo: DelegateMetaInfo{
+				HostName: t.c.delegateInfo.Host,
+				ID:       t.c.delegateInfo.ID,
+			},
+		},
+	}
+	httphelper.WriteJSON(w, resp, httpOK)
```
dhruv-harness, line 66: It would be easier to understand if we break the code in blocks eg

```
accountID := harness.GetAccountID(&req.Context, map[string]string{})

poolManager := t.c.getPoolManager(req.Distributed)
	
if !req.Distributed {
	harness.GetCtxState().Delete(req.StageRuntimeID)
}
	
err = harness.HandleDestroy(ctx, req, poolManager.GetStageOwnerStore(), &t.c.env, poolManager, t.c.metrics)
if err != nil {
	t.c.metrics.ErrorCount.WithLabelValues(accountID, strconv.FormatBool(req.Distributed)).Inc()
	
         logr.WithError(err).WithField("account_id", accountID).Error("could not destroy VM")
	
         httphelper.WriteJSON(w, failedResponse(err.Error()), httpFailed)
		
        return
}
```

types/types.go
```diff
@@ -42,29 +42,31 @@ const (
...
+	CAKey              []byte `db:"instance_ca_key" json:"ca_key"`
+	CACert             []byte `db:"instance_ca_cert" json:"ca_cert"`
+	TLSKey             []byte `db:"instance_tls_key" json:"tls_key"`
+	TLSCert            []byte `db:"instance_tls_cert" json:"tls_cert"`
+	Stage              string `db:"instance_stage" json:"stage"`
+	Updated            int64  `db:"instance_updated" json:"updated"`
+	Started            int64  `db:"instance_started" json:"started"`
+	IsHibernated       bool   `db:"is_hibernated" json:"is_hibernated"`
+	Port               int64  `db:"instance_port" json:"port"`
+	GitspacesAgentPort int64  `json:"gitspaces_agent_port"`
```
dhruv-harness, line 67: Please see my comment on the PR for one more port for vs code web and the ability to get additional ports during runtime or while starting containers/vm

dhruv-harness (commented):
Spacing between the code blocks can be improved quite a bit.

command/harness/dlite/gitspaces_cleanup.go
vikyathharekal: Most of the code in gitspaces_cleanup.go is borrowed from CI's cleanup.go. Same with gitspaces_init.go and init.go.
We have only changed the necessary things. 
gitspaces_init.go -> Request, Response, the additional parameter passed to HandleSetup method
gitspaces_cleanup.go -> Response

vikyathharekal: IMO We can move duplicate things to common code and make it more modular later. And that's when we can take up all the renaming items.
For now the plan is to make it work while not causing any regression in CI

n00bitax: yeah lets duplicate till we have complete flow, working and then come to refactoring, we don't have bandwidth to handle regression in existing flow.

dhruv-harness: sounds good @vikyathharekal

dhruv-harness (approved):

n00bitax (approved):

internal/drivers/nomad/driver.go
```diff
@@ -335,82 +351,86 @@ func (p *config) resourceJob(cpus, memGB int, vm, accountID string) (job *api.Jo
...
-		hostPath,
-		vmPath)
+	var runCmdFormat string
+	runCmdFormat = "%s run %s --name %s --cpus %s --memory %sGB --size %s --ssh --runtime=docker --ports %d:%s"
+	args := []interface{}{ignitePath, p.vmImage, vm, resource.Cpus, resource.MemoryGB, resource.DiskSize, liteEnginePort, strconv.Itoa(lehelper.LiteEnginePort)}
+	for vmPort, hostPort := range gitspacesPortMappings {
+		runCmdFormat += " --ports %d:%d"
+		args = append(args, hostPort, vmPort)
+	}
+	runCmdFormat += " --copy-files %s:%s"
```
vistaarjuneja, line 431: you can move this on L425

internal/drivers/nomad/driver.go
```diff
@@ -335,82 +351,86 @@ func (p *config) resourceJob(cpus, memGB int, vm, accountID string) (job *api.Jo
...
 
-	return ip, nodeID, port, nil
+	return ip, nodeID, ports, nil
 }
 
 // initJob creates a job which is targeted to a specific node. The job does the following:
 //  1. Starts a VM with the provided config
 //  2. Runs a startup script inside the VM
-func (p *config) initJob(vm, startupScript string, hostPort int, nodeID string, resource cf.NomadResource) (job *api.Job, id, group string) {
+func (p *config) initJob(vm, startupScript string, liteEnginePort int, gitspacesPortMappings map[int]int, nodeID string, resource cf.NomadResource) (job *api.Job, id, group string) {
```
vistaarjuneja, line 416: Please test CI once to make sure there's no regression on the existing flow

vistaarjuneja, line 416: Nit: `liteEnginePort` -> `liteEngineHostPort`

types/types.go
```diff
@@ -93,6 +94,7 @@ type InstanceCreateOpts struct {
 	AccountID            string
 	IsHosted             bool
 	ResourceClass        string
+	GitspaceOpts
```
vistaarjuneja, line 97: Let's not embed this into the struct - keep this `GitspaceOpts GitspaceOpts`

command/harness/dlite/gitspaces_cleanup.go
vistaarjuneja: I would push for removing duplication in this PR itself unless there's a very strong urgency, otherwise changes can very easily be missed for gitspaces.

Also I see two approaches here - one for adding gitspaces config in VM init/exec flows and the other (which you have done) is to create a separate task  and add in the new config required to those task structs. IMO the first approach would lead to less duplication and you could even keep the handler codes same. We could accept a gitspaces config in `SetupVMRequest` and logically speaking setting up gitspaces is akin to setting up a VM with some extra config. Was there any reason why you went with the second approach instead?

command/harness/dlite/routes.go
```diff
@@ -5,19 +5,23 @@ import (
...
+	initTask             = "DLITE_CI_VM_INITIALIZE_TASK"
+	executeTask          = "DLITE_CI_VM_EXECUTE_TASK"
+	executeTaskV2        = "DLITE_CI_VM_EXECUTE_TASK_V2"
+	cleanupTask          = "DLITE_CI_VM_CLEANUP_TASK"
+	cleanupTaskV2        = "DLITE_CI_VM_CLEANUP_TASK_V2"
+	gitspacesInitTask    = "DLITE_CDE_VM_INITIALIZE_TASK"
+	gitspacesCleanupTask = "DLITE_CDE_VM_CLEANUP_TASK"
 )
 
 func routeMap(c *dliteCommand) map[string]task.Handler {
```
vistaarjuneja, line 17: Do we need separate tasks for gitspaces? Why not re-use CI tasks?

internal/cloudinit/cloudinit.go
```diff
@@ -30,6 +30,7 @@ type Params struct {
 	PluginBinaryURI      string
 	Tmate                types.Tmate
 	IsHosted             bool
+	types.GitspaceAgentConfig
```
vistaarjuneja, line 33: let's not embed the struct

internal/drivers/manager.go
```diff
@@ -569,6 +569,11 @@ func (m *Manager) setupInstance(ctx context.Context, pool *poolEntry, tlsServerN
 	createOptions.Tmate = m.tmate
 	createOptions.AccountID = ownerID
 	createOptions.ResourceClass = resourceClass
+	if agentConfig != nil {
+		createOptions.Secret = agentConfig.Secret
```
vistaarjuneja, line 573: Let's not embed - this should look like `createOptions.GitOpts.Secret`

command/harness/dlite/gitspaces_cleanup.go
vistaarjuneja: @vikyathharekal @n00bitax @harshjain12 ^

internal/drivers/manager.go
```diff
@@ -306,7 +306,7 @@ func (m *Manager) StartInstancePurger(ctx context.Context, maxAgeBusy, maxAgeFre
 
 // Provision returns an instance for a job execution and tags it as in use.
 // This method and BuildPool method contain logic for maintaining pool size.
-func (m *Manager) Provision(ctx context.Context, poolName, runnerName, serverName, ownerID, resourceClass string, env *config.EnvConfig, query *types.QueryParams) (*types.Instance, error) {
+func (m *Manager) Provision(ctx context.Context, poolName, runnerName, serverName, ownerID, resourceClass string, env *config.EnvConfig, query *types.QueryParams, gitspaceAgentConfig *types.GitspaceAgentConfig) (*types.Instance, error) {
```
vistaarjuneja, line 309: This function is tied with pooling logic today (not ideal but it's been there since the beginning). In case of GCP/AWS, we keep a pool of VMs ready. This method selects a VM from that pool and adds another in the pool to recuperate that.
In hosted for nomad specifically we don't have a pool but we keep pools for GCP/AWS in some cases.
Let's say if we add pooling back for nomad, this function would behave in a weird way since it's checking for the pool but adding a non-gitspace VM in the pool.

Let's dissociate gitspaces from pooling completely so that each time a request comes in, we create a new VM. We also need to make sure gitspaces only uses the nomad pool since the implementation is only present there. For gitspaces, let's just call the setup directly in a separate function so we can keep it very simple and free from errors.

Something like this in the beginning of the function:
```
m.runnerName = runnerName
m.liteEnginePath = env.LiteEngine.Path
m.tmate = types.Tmate(env.Tmate)
pool := m.poolMap[poolName]
if pool == nil {
  return nil, fmt.Errorf("provision: pool name %q not found", poolName)
}
if pool.Driver != "nomad" { // double check the name
  return nil, fmt.Errorf("incorrect pool, gitspaces is only supported on nomad")
}
if gitspaceAgentConfig != nil {
   inst, err = m.setupInstance(ctx, pool, serverName, ownerID, resourceClass, true, gitspaceAgentConfig)
   return inst, err
}
```

internal/drivers/nomad/driver.go
```diff
@@ -335,82 +351,84 @@ func (p *config) resourceJob(cpus, memGB int, vm, accountID string) (job *api.Jo
...
+	}
 
 	// sanity check
-	if port <= 0 || port > 65535 {
-		err = fmt.Errorf("scheduler: port %d generated is not a valid port", port)
-		logr.Errorln(err)
-		return ip, nodeID, port, err
+	for _, port := range ports {
+		if port <= 0 || port > 65535 {
+			err = fmt.Errorf("scheduler: port %d generated is not a valid ports", port)
```
vistaarjuneja, line 391: Nit: ports -> port in log message

internal/drivers/nomad/driver.go
```diff
@@ -335,82 +351,84 @@ func (p *config) resourceJob(cpus, memGB int, vm, accountID string) (job *api.Jo
...
 
-	return ip, nodeID, port, nil
+	return ip, nodeID, ports, nil
 }
 
 // initJob creates a job which is targeted to a specific node. The job does the following:
 //  1. Starts a VM with the provided config
 //  2. Runs a startup script inside the VM
-func (p *config) initJob(vm, startupScript string, hostPort int, nodeID string, resource cf.NomadResource) (job *api.Job, id, group string) {
+func (p *config) initJob(vm, startupScript string, liteHostEnginePort int, gitspacesPortMappings map[int]int, nodeID string, resource cf.NomadResource) (job *api.Job, id, group string) {
```
vistaarjuneja, line 416: Nit: `liteHostEnginePort` -> `liteEngineHostPort`

internal/drivers/nomad/driver.go
```diff
@@ -185,39 +185,55 @@ func (p *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (*t
 	logr.Infoln("scheduler: found a node with available resources")
 
 	// get the machine details where the resource job was allocated
-	ip, id, hostPort, err := p.fetchMachine(logr, resourceJobID)
+	ip, id, hostPorts, err := p.fetchMachine(logr, resourceJobID)
 	if err != nil {
 		defer p.deregisterJob(logr, resourceJobID, false) //nolint:errcheck
 		return nil, err
 	}
+	liteHostEnginePort := hostPorts[vm]
```
vistaarjuneja, line 193: Nit: liteEngineHostPort

vistaarjuneja, line 193: Let's handle error appropriately in case `hostPorts` does not have en entry to get `liteEngineHostPort`.
You can have the fetchMachine function return lite engine host port separately and gitspaces ports separately.
`ip, id, liteEngineHostPort, gitspacesPorts, err := ...`
Also, do we need hostPorts to be a map here? Can't we just return `gitspacesPorts` as a list of integers?

internal/drivers/nomad/driver.go
```diff
@@ -768,3 +789,11 @@ func generateDestroyCommand(vm string) string {
 		fi
 	`, ignitePath, vm, ignitePath, vm, ignitePath, vm, ignitePath, vm)
 }
+
+func getNetworkResources(portLabel string, ports []int) []*api.NetworkResource {
```
vistaarjuneja, line 793: Add a comment here as to what this function does. 
And also comment about why the port label is of the specific format

internal/drivers/nomad/driver.go
vikyathharekal: liteEngineHostPort can be returned separately. 
Actually we just need x number of ports. We don't even have to send the list ports to resourceJob() method.
And yes fetchMachine can return just a list of integers. We can then map it appropriately before sending it to the initJob() where we construct the runCmd.

internal/drivers/nomad/driver.go
vikyathharekal: I'll be changing this. We can have the labels as `<vmid>_gitspaces_1`, `<vmid>_gitspaces_2` etc. We don't need specific port names

internal/drivers/nomad/driver.go
```diff
@@ -768,3 +794,15 @@ func generateDestroyCommand(vm string) string {
 		fi
 	`, ignitePath, vm, ignitePath, vm, ignitePath, vm, ignitePath, vm)
 }
+
+// Request Nomad to assign available ports dynamically.
+// 1 for lite engine and another n ports for gitspaces as requested
+// Since the port labels have to be unique, VM ID is used usually.
+// For gitspaces port, prefix of _gitspaces_{port_number_count} is added to the label to keep it unique
+func getNetworkResources(portLabel string, portsCount int) []*api.NetworkResource {
```
vistaarjuneja, line 802: Nit: could we rename the fields to make it more explicit what is being passed here?
`portLabel` -> `liteEnginePortLabel`
`portsCount` -> `gitspacesPortCount`
Also - just curious - what do we use the multiple git spaces ports for?

internal/drivers/nomad/driver.go
```diff
@@ -335,82 +351,89 @@ func (p *config) resourceJob(cpus, memGB int, vm, accountID string) (job *api.Jo
...
-	port = alloc.Resources.Networks[0].DynamicPorts[0].Value
+	liteEngineHostPort = alloc.Resources.Networks[0].DynamicPorts[0].Value
+	for i := 1; i < len(alloc.Resources.Networks[0].DynamicPorts); i++ {
+		ports = append(ports, alloc.Resources.Networks[0].DynamicPorts[i].Value)
+	}
 
 	// sanity check
-	if port <= 0 || port > 65535 {
-		err = fmt.Errorf("scheduler: port %d generated is not a valid port", port)
+	if liteEngineHostPort <= 0 || liteEngineHostPort > 65535 {
```
vistaarjuneja, line 389: Nit: just combine both these into a single loop after iterating over ports and keep the error message
fmt.Errorf("scheduler: port %d generated is not a valid port", port)

internal/drivers/nomad/driver.go
vikyathharekal: It's not just liteEnginePortLabel right? It's used for the other 2 ports as well
We need one port for the gitspaces agent
And second port for either SSH or web (depending on the IDE type chosen vscode/vscode web)

internal/drivers/nomad/driver.go
vikyathharekal: I didn't get it, how do you combine it into a single loop? 
One option is to iterate over alloc.Resources.Networks[0].DynamicPorts before we assign liteEngineHostPort and the ports slice/array

vistaarjuneja: ah ok - this is fine, we can keep as is

vistaarjuneja (approved):

internal/drivers/nomad/driver.go
vikyathharekal: Ran automation locally. There are a few failures (github connector issue), but these are failing even on the main branch (without my changes). cc: @harshjain12

vikyathharekal:
trigger messagecheck

internal/drivers/nomad/driver.go
```diff
@@ -185,39 +185,55 @@ func (p *config) Create(ctx context.Context, opts *types.InstanceCreateOpts) (*t
...
 		defer p.deregisterJob(logr, resourceJobID, false) //nolint:errcheck
 		return nil, err
 	}
 
+	var gitspacesPortMappingsString string
+	gitspacesPortMappings := make(map[int]int)
+
+	if len(opts.GitspaceOpts.Ports) != len(gitspacesPorts) {
+		return nil, fmt.Errorf("scheduler: could not allocate required number of ports for gitspaces")
+	} else {
```
harshjain12, line 199: this will run for non git space ?

internal/drivers/nomad/driver.go
```diff
@@ -335,82 +351,89 @@ func (p *config) resourceJob(cpus, memGB int, vm, accountID string) (job *api.Jo
...
-		err = fmt.Errorf("scheduler: port %d generated is not a valid port", port)
+	if liteEngineHostPort <= 0 || liteEngineHostPort > 65535 {
+		err = fmt.Errorf("scheduler: lite engine host port %d generated is not a valid port", liteEngineHostPort)
 		logr.Errorln(err)
-		return ip, nodeID, port, err
+		return ip, nodeID, liteEngineHostPort, ports, err
+	}
+	for _, port := range ports {
+		if port <= 0 || port > 65535 {
+			err = fmt.Errorf("scheduler: gitspace host port %d generated is not a valid port", port)
```
harshjain12, line 396: How are we making sure that this will not called for bare metal ?

internal/drivers/nomad/driver.go
```diff
@@ -335,82 +351,89 @@ func (p *config) resourceJob(cpus, memGB int, vm, accountID string) (job *api.Jo
...
-		resource.MemoryGB,
-		resource.DiskSize,
-		hostPort,
-		strconv.Itoa(lehelper.LiteEnginePort),
-		hostPath,
-		vmPath)
+	var runCmdFormat string
+	runCmdFormat = "%s run %s --name %s --cpus %s --memory %sGB --size %s --ssh --runtime=docker --ports %d:%s --copy-files %s:%s"
+	args := []interface{}{ignitePath, p.vmImage, vm, resource.Cpus, resource.MemoryGB, resource.DiskSize, liteEngineHostPort, strconv.Itoa(lehelper.LiteEnginePort), hostPath, vmPath}
+	for vmPort, hostPort := range gitspacesPortMappings {
```
harshjain12, line 432: how are we making sure that it will be called only for CDE ?

internal/drivers/nomad/driver.go
vikyathharekal: For non-gitspace, both `opts.GitspaceOpts.Ports` and `gitspacesPorts` will be nil. 
Nil slices have length of 0, so it will go to the else block.
In the else block, since the length of opts.GitspaceOpts.Ports is 0, it will not run the for loop

internal/drivers/nomad/driver.go
vikyathharekal: For non-gitspace, ports will be a nil slice with length 0, hence the for loop will not be executed.

internal/drivers/nomad/driver.go
vikyathharekal: For non-gitspace, gitspacesPortMappings is an empty map, execution will not go into the for loop

PR #509, opened 2024-10-03
feat: [CI-14261]: Added support of hardware acceleration in GCP VMs

ShobhitSingh11:
# Commit Checklist

## The Basics

### DIFF
--- a/store/database/migrate/postgres/0011_update_table_instances_nested_virtualization.up.sql
+++ b/store/database/migrate/postgres/0011_update_table_instances_nested_virtualization.up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE instances ADD COLUMN IF NOT EXISTS enable_nested_virtualization BOOLEAN NOT NULL DEFAULT false;
--- a/command/config/config.go
+++ b/command/config/config.go
@@ -190,23 +190,24 @@ type (
 
 	// Google specifies the configuration for a GCP instance.
 	Google struct {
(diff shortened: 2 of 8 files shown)


store/database/migrate/postgres/0011_update_table_instances_nested_virtualization.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE instances ADD COLUMN IF NOT EXISTS enable_nested_virtualization BOOLEAN NOT NULL DEFAULT false;
```
raghavharness, line 1: do we need to store in database ? why ?

ShobhitSingh11: Reason is we need to log during init whether nedtedVirt is enabled or not, for that we are storing nestedVirt variable in instance property and use this instance property to show it in log. For pools > 0 without storing it in db that instance variable is not populating properly (was always showing false, even if nestedVirt is actually enabled in instance). However instance creation was working properly according to the property set in pool.yml

raghavharness (commented):
Looks good, please add in the grafana metrics as well

store/database/migrate/postgres/0011_update_table_instances_nested_virtualization.up.sql
raghavharness: add enable_nested_virtualization property in grafana metrics also

raghavharness (approved):

PR #510, opened 2024-10-04
feat: [CDE-365]: Send the whole gitspace agent script instead of individual params

vikyathharekal:
# Description
Since we wanted to send more params for container name, tag, image etc, instead of adding so many params, thought of sending the whole script and appending to the existing one in cloudinit.go
To keep it backward compatible, I haven't removed any existing params.
This would be the script coming from CDE Manager with all the replaced values.
```
groupadd docker
mkdir -p /mnt/disks/mountdevcontainer/gitspace
export DOCKER_API_VERSION={{ .AgentDockerAPIVersion }}

echo "Updating docker root dir"
systemctl stop docker
mkdir -p /mnt/disks/mountdevcontainer/docker
tee /etc/docker/daemon.json <<EOF
{
  "data-root": "/mnt/disks/mountdevcontainer/docker"
}
EOF
systemctl start docker
echo "Successfully updated docker root dir"

echo "starting gitspaces agent"

# Check if the container is present

if docker ps -a --filter "name={{ .AgentContainerName }}" --quiet | grep -q .; then
    echo "Container {{ .AgentContainerName }} exists. Attempting to start it..."

    # Try to start the container
    docker start {{ .AgentContainerName }}
    if [ $? -ne 0 ]; then
        echo "Failed to start {{ .AgentContainerName }}. Deleting and recreating..."
        docker rm -f {{ .AgentContainerName }}
        echo {{ .AgentContainerRegistryToken }} | docker login -u oauth2accesstoken --password-stdin {{ .AgentContainerRegistry }}
        docker run -p 8083:8083 -v /var/run/docker.sock:/var/run/docker.sock -e DOCKER_API_VERSION={{ .AgentDockerAPIVersion }} -e HARNESS_JWT_SECRET={{ .AgentSecret }} -d --name {{ .AgentContainerName }} {{ .AgentImage }}:{{ .AgentTag }}
    else
        echo "{{ .AgentContainerName }} started successfully."
    fi
else
    echo "Container {{ .AgentContainerName }} does not exist. Creating it..."
    echo {{ .AgentContainerRegistryToken }} | docker login -u oauth2accesstoken --password-stdin {{ .AgentContainerRegistry }}
    docker run -p 8083:8083 -v /var/run/docker.sock:/var/run/docker.sock -e DOCKER_API_VERSION={{ .AgentDockerAPIVersion }} -e HARNESS_JWT_SECRET={{ .AgentSecret }} -d --name {{ .AgentContainerName }} {{ .AgentImage }}:{{ .AgentTag }}
    echo "Agent container started."
fi

echo "done starting gitspaces agent"
```

# Commit Checklist

## The Basics

### DIFF
--- a/internal/cloudinit/cloudinit.go
+++ b/internal/cloudinit/cloudinit.go
@@ -250,6 +250,20 @@ echo "done starting lite engine server"
 {{ if .GitspaceAgentConfig.VMInitScript }}
 {{ .GitspaceAgentConfig.VMInitScript }}
 {{ else }}
+groupadd docker
+mkdir -p /opt/gitspaceagent
+
+echo "Updating docker root dir"
+systemctl stop docker
+mkdir -p /mnt/disks/mountdevcontainer/docker
+tee /etc/docker/daemon.json <<EOF
+{
+  "data-root": "/mnt/disks/mountdevcontainer/docker"
+}
+EOF
+systemctl start docker
+echo "Successfully updated docker root dir"
+
 echo "downloading gitspaces agent binary"
 echo HARNESS_JWT_SECRET={{ .GitspaceAgentConfig.Secret }} >> /etc/profile
 export HARNESS_JWT_SECRET={{ .GitspaceAgentConfig.Secret }}
--- a/internal/cloudinit/cloudinit.go
+++ b/internal/cloudinit/cloudinit.go
@@ -388,8 +388,8 @@ func LinuxBash(params *Params) (payload string) {
 	if (params.GitspaceAgentConfig.Secret != "" && params.GitspaceAgentConfig.AccessToken != "") ||
 		(params.GitspaceAgentConfig.VMInitScript != "") {
 		if params.GitspaceAgentConfig.VMInitScript != "" {
-			decodedScript, err := base64.StdEncoding.DecodeString(params.GitspaceAgentConfig.VMInitScript)
-			if err != nil {
+			decodedScript, decodeErr := base64.StdEncoding.DecodeString(params.GitspaceAgentConfig.VMInitScript)
+			if decodeErr != nil {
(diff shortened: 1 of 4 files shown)


internal/cloudinit/cloudinit.go
```diff
@@ -395,7 +399,16 @@ func LinuxBash(params *Params) (payload string) {
...
 
 	var err error
-	if params.GitspaceAgentConfig.Secret != "" && params.GitspaceAgentConfig.AccessToken != "" {
+	if (params.GitspaceAgentConfig.Secret != "" && params.GitspaceAgentConfig.AccessToken != "") ||
+		(params.GitspaceAgentConfig.VMInitScript != "") {
+		if params.GitspaceAgentConfig.VMInitScript != "" {
+			decodedScript, decodeErr := base64.StdEncoding.DecodeString(params.GitspaceAgentConfig.VMInitScript)
+			if decodeErr != nil {
+				err = fmt.Errorf("failed to decode the gitspaces vm init script: %w", err)
+				panic(err)
```
dhruv-harness, line 408: Should we panic here? How about we return the error?

vikyathharekal: That's the convention followed throughout this file, so I continued doing so 🙂

vistaarjuneja: I agree that we shouldn't panic (even though I believe we handle stacktraces and don't crash the program but I may be wrong). Can you change this (and other instances of panic if possible otherwise we can take that up later). The error can be propagated up

vikyathharekal: Sure I can do it. It would mean changes in a lot of files. I would prefer to take it separately in a different PR

vikyathharekal: Around 14 files

vistaarjuneja (approved):

PR #535, opened 2025-01-21
feat: [CDE-583]: Make destroy operation stateless by passing instance details instead of relying on db / in memory storage

vikyathharekal:
# Description
- Passing zone in create/setup requests. If present, a random zone won't be selected. (CI requests will continue with random zone selection)
- This instance info is saved in the DB of cde-manager and passed in the destroy request.
- Destroy requests will fetch instances from DB only when not present in the request. (CI requests will continue to fetch )

# Testing
- CDE requests use the passed information from the destroy requests instead of relying on the instance record in DB
- CI requests continue to use the instance stored in DB in the destroy/cleanup request
Screenshot 2025-01-21 at 12 56 23 PM

# Commit Checklist

## The Basics

### DIFF
--- a/command/harness/helper.go
+++ b/command/harness/helper.go
@@ -1,8 +1,12 @@
 package harness
 
 import (
+	"errors"
+	"fmt"
+	"github.com/drone-runners/drone-runner-aws/types"
 	"hash/fnv"
 	"path/filepath"
+	"reflect"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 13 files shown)


command/harness/helper.go
```diff
@@ -35,3 +39,43 @@ func fileID(filename string) string {
 	h.Write([]byte(filename))
 	return strings.Replace(filepath.Base(filename), ".", "-", -1) + strconv.Itoa(int(h.Sum32()))
 }
+
+// validateStruct checks if all fields of a struct are populated.
+func validateStruct(data interface{}) error {
```
vistaarjuneja, line 44: Is there an alternative to using reflection here?

vikyathharekal: Alternative is to check each field manually

command/harness/helper.go
```diff
@@ -35,3 +39,43 @@ func fileID(filename string) string {
...
+	v := reflect.ValueOf(data)
+
+	// Ensure the input is a struct
+	if v.Kind() != reflect.Struct {
```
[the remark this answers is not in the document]
vikyathharekal: reflect.Value is not a pointer and therefore cannot be directly compared to nil

vistaarjuneja (approved):

PR #538, opened 2025-01-28
feat: [CDE-585]: Support creation and attaching of persistent disks

vikyathharekal:
# Description
1. Create Disk
    - Create (if required) and attach the persistent disk to the VM.
    - Wait for the disk to be created and then initiate instance creation.
    - Pass storage type (example: "pd-balanced")
    - storage_identifier supports a comma-separated list of identifiers, to support 2 separate disks (1 for gitspace and 1 for docker) in the future.
2. Delete Disk
    - Delete the disk on Delete Gitspace (StorageCleanupType.DELETE).
    - Disk would retained in case of Stop Gitspace (StorageCleanupType.DETACH).
    - If the disk needs to be deleted, wait for instance deletion to complete and then hit the delete disk API.
    - Add storage_identifier in instance info (This will be sent in the destroy request to identify the disk and delete it)
    - Pass StorageCleanupType in VMCleanupRequest
3. Skip CDE VMs from cleanup by setting "retain": true label

# Commit Checklist

## The Basics

### DIFF
--- a/app/drivers/google/driver.go
+++ b/app/drivers/google/driver.go
@@ -371,7 +371,10 @@ func (p *config) attachPersistentDisk(
 		if err != nil {
 			return nil, err
 		}
-		operations = append(operations, op)
+		if op != nil {
+			// this means we have submitted disk creation request(s)
+			operations = append(operations, op)
+		}
 
 		// attach to instance
 		attachedDisk := &compute.AttachedDisk{
--- a/app/drivers/google/driver.go
+++ b/app/drivers/google/driver.go
@@ -299,8 +299,8 @@ func (p *config) create(ctx context.Context, opts *types.InstanceCreateOpts, nam
 	if opts.StorageOpts.Identifier != "" {
 		operations, attachDiskErr := p.attachPersistentDisk(ctx, opts, in, zone)
 		if attachDiskErr != nil {
-			logr.WithError(err).Errorln("google: failed to attach persistent disk")
-			return nil, err
+			logr.WithError(attachDiskErr).Errorln("google: failed to attach persistent disk")
+			return nil, attachDiskErr
 		}
 		for _, operation := range operations {
 			if operation != nil {
--- a/app/drivers/google/driver.go
+++ b/app/drivers/google/driver.go
@@ -595,11 +595,7 @@ func (p *config) resumeInstance(ctx context.Context, projectID, zone, name strin
 
 func (p *config) insertInstance(ctx context.Context, projectID, zone, requestID string, in *compute.Instance) (*compute.Operation, error) {
 	return retry(ctx, insertRetries, secSleep, func() (*compute.Operation, error) {
-		op, err := p.service.Instances.Insert(projectID, zone, in).RequestId(requestID).Context(ctx).Do()
-		if err != nil {
-			return op, err
-		}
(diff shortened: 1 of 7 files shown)


raghavharness (approved):
LGTM

app/drivers/google/driver.go
```diff
@@ -394,21 +459,52 @@ func (p *config) DestroyInstanceAndStorage(ctx context.Context, instances []*typ
...
 
-		requestID := uuid.New().String()
-
-		_, err = p.deleteInstance(ctx, p.projectID, zone, instance.ID, requestID)
-		if err != nil {
+		instanceDeleteOperation, deleteInstanceErr := p.deleteInstance(ctx, p.projectID, zone, instance.ID, uuid.New().String())
+		if deleteInstanceErr != nil {
 			// https://github.com/googleapis/google-api-go-client/blob/master/googleapi/googleapi.go#L135
-			if gerr, ok := err.(*googleapi.Error); ok &&
+			if gerr, ok := deleteInstanceErr.(*googleapi.Error); ok &&
```
raghavharness, line 465: this would be a behaviour change, since earlier we were returning `err` and now changed the variable name to `deleteInstanceErr`.
You can assign err = deleteInstanceErr after `if deleteInstanceErr != nil` check

vikyathharekal: Fixed

app/drivers/google/driver.go
```diff
@@ -295,13 +296,29 @@ func (p *config) create(ctx context.Context, opts *types.InstanceCreateOpts, nam
 		}
 	}
 
-	requestID := uuid.New().String()
-	op, err := p.insertInstance(ctx, p.projectID, zone, requestID, in)
+	if opts.StorageOpts.Identifier != "" {
+		operations, attachDiskErr := p.attachPersistentDisk(ctx, opts, in, zone)
+		if attachDiskErr != nil {
+			logr.WithError(err).Errorln("google: failed to attach persistent disk")
+			return nil, err
```
raghavharness, line 303: should be attachDiskErr ?

app/drivers/google/driver.go
```diff
@@ -394,21 +459,53 @@ func (p *config) DestroyInstanceAndStorage(ctx context.Context, instances []*typ
...
+				)
+				if diskDeletionErr != nil {
+					var googleErr *googleapi.Error
+					if errors.As(diskDeletionErr, &googleErr) &&
+						googleErr.Code == http.StatusNotFound {
+						logr.WithError(diskDeletionErr).
+							Errorln("google: persistent disk %s not found", storageIdentifier)
+					}
+				}
+				err = p.waitZoneOperation(ctx, diskDeleteOperation.Name, zone)
```
raghavharness, line 500: Can diskDeleteOperation be nil if diskDeletionErr != nil ?

vikyathharekal: It should not be. We are doing similar operations already for instance creation and deletion. Only in case of createPersistentDiskIfNotExists we are doing custom handling were operation can be null

app/drivers/google/driver.go
vikyathharekal: fixed

harshjain12 (approved):

PR #556, opened 2025-03-04
feat: [CDE-616]: Don't log the whole response which now contains certs and keys

vikyathharekal:
Screenshot 2025-03-04 at 10 33 31 AM

# Commit Checklist

## The Basics

### DIFF
--- a/command/harness/setup.go
+++ b/command/harness/setup.go
@@ -215,7 +215,6 @@ func HandleSetup(
 		WithField("ip", instance.Address).
 		WithField("id", instance.ID).
 		WithField("instance_name", instance.Name).
-		WithField("response", fmt.Sprintf("%+v", resp)).
 		WithField("tried_pools", pools).
 		Traceln("VM setup is complete")
 


devkimittal:
screenshot of init logs of both dlite and runner

command/harness/setup.go
```diff
@@ -215,7 +215,6 @@ func HandleSetup(
 		WithField("ip", instance.Address).
 		WithField("id", instance.ID).
 		WithField("instance_name", instance.Name).
-		WithField("response", fmt.Sprintf("%+v", resp)).
```
devkimittal, line 218: can you please point me to the pr where this was added?

vikyathharekal: Not sure. I haven't added it, it must be there from a long time

vikyathharekal:
> screenshot of init logs of both dlite and runner

I have added screenshot from my IDE. It would print the same in the init step.

command/harness/setup.go
devkimittal: why not just revert the change that was printing the extra info

vikyathharekal: The log line to log the response was there from a long time I guess. Earlier the response had only instance id and IP. Now it has a lot more things required for CDE

We still need those information to be returned for CDE, including the certs and keys.
Just that we don't have to log them

vikyathharekal: Since we already log instance id and IP above it
```
		WithField("ip", instance.Address).
		WithField("id", instance.ID).
```
I have removed the response part as it would have redundant info + info required for CDE

devkimittal (approved):

PR #561, opened 2025-03-18
feat: [CDE-666]: Support suspend and resume for gitspaces

vikyathharekal:
# Description
- Add compile-time checks to ensure all implementations satisfy the drivers.Driver interface.
- Pass instance info in start() request so that if the VM is already present in suspended state, it can be resumed
- Pass the zone to the hibernate method. If empty, the fallback would be to check the DB, which is the current behaviour.
- Add /suspend API. The delegate would call this in case it processes VMSuspendTask
- Suspend handler builds the instance from instance info passed in the request and suspends the VM. It further calls LE /suspend.
- Make boot disk type and size configurable for CDE

# Commit Checklist

## The Basics


go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.3.0
 	github.com/google/wire v0.5.0
-	github.com/harness/lite-engine v0.5.103
+	github.com/harness/lite-engine v1.0.0-rc.137.0.20250407043243-5794fe017d1d
```
raghavharness, line 27: Use the correct version of LE when that PR is merged.

vikyathharekal: Updated

raghavharness (approved):

PR #568, opened 2025-03-24
fix: [CI-16185]: Force delete firecracker vm

raghavharness:
Force delete firecracker vm

### DIFF
--- a/app/drivers/nomad/linux_virtualizer.go
+++ b/app/drivers/nomad/linux_virtualizer.go
@@ -290,7 +290,11 @@ func (lv *LinuxVirtualizer) GetDestroyScriptGenerator() func(string, string) str
 		if [ $? -ne 0 ]; then
 		  %s stop -f %s; %s rm -f %s
 		fi
-	`, ignitePath, vm, ignitePath, vm, ignitePath, vm, ignitePath, vm)
+		vm_id=$(ignite ps | grep %s | awk '{print $1}')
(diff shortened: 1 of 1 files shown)


app/drivers/nomad/linux_virtualizer.go
```diff
@@ -290,7 +290,11 @@ func (lv *LinuxVirtualizer) GetDestroyScriptGenerator() func(string, string) str
 		if [ $? -ne 0 ]; then
 		  %s stop -f %s; %s rm -f %s
 		fi
-	`, ignitePath, vm, ignitePath, vm, ignitePath, vm, ignitePath, vm)
+		vm_id=$(ignite ps | grep %s | awk '{print $1}')
```
HenAmar, line 293: Is that the only way to handle it? What about the case you mentioned where line 291 goes out of Nomad execution?

raghavharness: I was wrong about it, it goes not next line even if it fails. So for now forcefully removing firecracker vm from disk.

HenAmar: do we have a plan to figure this out, again i assume we get cni error, do you think this is the case?
image

raghavharness: Removing the firecracker VM all together should help with this no ?

HenAmar (approved):

PR #572, opened 2025-04-02
feat:[CI-16771]: Add tmate compiled for arm

HenAmar:
For the PayPal POV, we need to support remote debugging on Mac. This requires compiling tmate statically on Mac and pulling it via the CloudInit script.

### DIFF
--- a/app/cloudinit/cloudinit.go
+++ b/app/cloudinit/cloudinit.go
@@ -402,6 +402,21 @@ else
     fi
 fi
 chmod 777 /usr/local/bin/plugin
+
+{{ if .Tmate.Enabled }}
+mkdir /tmp/addon
+if wget -nv https://github.com/harness/tmate/releases/download/1.0/tmate-1.0-static-mac-arm64.tar.xz -O /tmp/addon/tmate.xz; then
+	echo "Successfully downloaded tmate binary from primary URL."
+else
+	echo "Primary URL failed for tmate. Trying fallback URL..."
+	wget -nv https://app.harness.io/storage/harness-download/harness-ti/harness-tmate/1.0/tmate-1.0-static-mac-arm64.tar.xz -O /tmp/addon/tmate.xz
+	echo "Successfully downloaded tmate binary from fallback URL."
+fi
+tar -xf /tmp/addon/tmate.xz -C /tmp/addon/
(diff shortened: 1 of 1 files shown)


app/cloudinit/cloudinit.go
```diff
@@ -402,6 +402,21 @@ else
     fi
 fi
 chmod 777 /usr/local/bin/plugin
+
+{{ if .Tmate.Enabled }}
```
devkimittal, line 406: is this the only place we need to make the change? for both tart and anka?

app/cloudinit/cloudinit.go
```diff
@@ -402,6 +402,21 @@ else
...
+mkdir /tmp/addon
+if wget -nv https://github.com/harness/tmate/releases/download/1.0/tmate-1.0-static-mac-arm64.tar.xz -O /tmp/addon/tmate.xz; then
+	echo "Successfully downloaded tmate binary from primary URL."
+else
+	echo "Primary URL failed for tmate. Trying fallback URL..."
+	wget -nv https://app.harness.io/storage/harness-download/harness-ti/harness-tmate/1.0/tmate-1.0-static-mac-arm64.tar.xz -O /tmp/addon/tmate.xz
```
Sapa96, line 412: can we sync this binary in GCS bucket ? i dont see it currently 
https://console.cloud.google.com/storage/browser/harness-ti/harness-tmate/1.0?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&invt=AbtrKA

HenAmar: @devkimittal done.

app/cloudinit/cloudinit.go
HenAmar: yes

HenAmar: anka is irrelevant and this is the only script for mac. The other mac is irrelevant as we have only arm64 (tart support only apple silicon)

Sapa96 (approved):

app/cloudinit/cloudinit.go
hemanthmantri: Nit: File extensions don't matter in Linux but it would be better to retain the extension of the original file when it's downloaded. `/tmp/addon/tmate.tar.xz`

PR #580, opened 2025-04-22
feat: [CI-16869]: HA for vm runner

devkimittal:
This pr attempts to support running multiple replicas of vm runner in a kubernetes cluster while delegate being a single one. We are introducing this flag DRONE_RUNNER_HA which should be enabled in case we are deploying multiple replicas.
Load tested with one delegate and 8 runner replicas on kubernetes https://app.harness.io/ng/account/9UuUfLwaQ-6ZowvbG7qtLQ/module/ci/orgs/default/projects/RaghavTest/pipelines/raghavnginxbuild/executions/CB1dIPBSSv2O-Dkegv0EhA/pipeline?stage=5te4Ng8xRzCb3fqeyN7dMg&stageExecId=W3feAM0ZTCCuUzFHtijzXA
https://app.harness.io/ng/account/9UuUfLwaQ-6ZowvbG7qtLQ/module/ci/orgs/default/projects/RaghavTest/pipelines/raghavnginxbuild/executions/PcC7PawGRJyvbMttJ54MxA/pipeline

### DIFF
--- a/app/drivers/manager.go
+++ b/app/drivers/manager.go
@@ -472,7 +472,8 @@ func (m *Manager) Destroy(ctx context.Context, poolName, instanceID string, inst
 }
 
 func (m *Manager) BuildPools(ctx context.Context) error {
-	return m.forEach(ctx, m.GetTLSServerName(), nil, m.buildPoolWithMutex)
+	query := types.QueryParams{RunnerName: m.runnerName}
+	return m.forEach(ctx, m.GetTLSServerName(), &query, m.buildPoolWithMutex)
 }
 
 func (m *Manager) cleanPool(ctx context.Context, pool *poolEntry, query *types.QueryParams, destroyBusy, destroyFree bool) error {
@@ -514,7 +515,7 @@ func (m *Manager) cleanPool(ctx context.Context, pool *poolEntry, query *types.Q
 
 func (m *Manager) CleanPools(ctx context.Context, destroyBusy, destroyFree bool) error {
 	var returnError error
-	query := types.QueryParams{MatchLabels: map[string]string{"retain": "false"}}
+	query := types.QueryParams{RunnerName: m.runnerName, MatchLabels: map[string]string{"retain": "false"}}
 	for _, pool := range m.poolMap {
 		err := m.cleanPool(ctx, pool, &query, destroyBusy, destroyFree)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


app/drivers/manager.go
```diff
@@ -67,6 +68,7 @@ func New(
 		autoInjectionBinaryURI:  env.Settings.AutoInjectionBinaryURI,
 		liteEngineFallbackPath:  env.LiteEngine.FallbackPath,
 		pluginBinaryFallbackURI: env.Settings.PluginBinaryFallbackURI,
+		env:                     env,
```
raghavharness, line 71: won't new runner break ?

command/harness/setup.go
```diff
@@ -245,11 +245,8 @@ func handleSetup(
 	stageRuntimeID := r.ID
 
 	// try to provision an instance from the pool manager.
-	var query *types.QueryParams
-	if poolManager.IsDistributed() {
-		query = &types.QueryParams{
-			RunnerName: runnerName,
-		}
```
raghavharness, line 252: you tested for non HA builds as well ?

raghavharness (changes requested):

command/harness/setup.go
devkimittal: non ha builds will have same runnerName in all instances, so query won't matter.

app/drivers/manager.go
devkimittal: ACK

command/harness/delegate/delegate.go
```diff
@@ -114,7 +114,7 @@ func (c *delegateCommand) run(*kingpin.ParseContext) error {
 	c.poolManager = drivers.New(ctx, instanceStore, &c.env)
 
 	_, err = harness.SetupPoolWithEnv(ctx, &c.env, c.poolManager, c.poolFile)
-	defer harness.Cleanup(c.env.Settings.ReusePool, c.poolManager, true, true) //nolint: errcheck
+	defer harness.Cleanup(c.env.Settings.ReusePool, c.poolManager, false, true) //nolint: errcheck
 	if err != nil {
```
raghavharness, line 118: this will also not destroy free vms for non-ha builds

devkimittal: this flag is for busyvms

raghavharness (approved):

PR #596, opened 2025-05-14
[feat]: [CI-17465]: reducing retry count to fail fast from github

Sapa96:
1. changes are done only in linux Script & gitspacesLinuxScript (Baremetal) for plugin and le binaries 
2. All the other places primary source download fails in 3 retries already 

Testing : 

tested by giving primary url wrong link and verified 3 times download retry 

Screenshot 2025-05-14 at 21 29 10
Screenshot 2025-05-14 at 21 29 20

tested primary source download post script changes 

Screenshot 2025-05-14 at 21 55 15

# Commit Checklist

## The Basics

### DIFF
--- a/app/cloudinit/cloudinit.go
+++ b/app/cloudinit/cloudinit.go
@@ -115,7 +115,7 @@ swapon /swapfile
 echo "done setting up swap space"
 
 echo "downloading lite engine binary"
-if /usr/bin/wget --retry-connrefused --retry-on-host-error --retry-on-http-error=503,404,429 --tries=10 --waitretry=10 ` + liteEngineUsrBinPath + ` || /usr/bin/wget --retry-connrefused --tries=10 --waitretry=10 -nv --debug ` + liteEngineUsrBinPath + `; then
+if /usr/bin/wget --retry-connrefused --retry-on-host-error --retry-on-http-error=503,404,429 --tries=3 --waitretry=3 ` + liteEngineUsrBinPath + ` || /usr/bin/wget --retry-connrefused --tries=3 --waitretry=3 -nv --debug ` + liteEngineUsrBinPath + `; then
     echo "Successfully downloaded lite engine binary from primary URL."
 else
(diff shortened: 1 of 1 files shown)


Sapa96:
trigger messagecheck

Sapa96:
trigger messageCheck

devkimittal:
@ShobhitSingh11 added this because the download command was failing for a particular version on vm.

app/cloudinit/cloudinit.go
```diff
@@ -128,7 +128,7 @@ cp "/etc/environment" $HOME/.env
 echo "SKIP_PREPARE_SERVER=true" >> $HOME/.env;
 
 {{ if .PluginBinaryURI }}
-if wget --retry-connrefused --retry-on-host-error --retry-on-http-error=503,404,429 --tries=10 --waitretry=10 ` + pluginUsrBinPath + ` || wget --retry-connrefused --tries=10 --waitretry=10 ` + pluginUsrBinPath + `; then
+if wget --retry-connrefused --retry-on-host-error --retry-on-http-error=503,404,429 --tries=3 --waitretry=3 ` + pluginUsrBinPath + `; then
```
ShobhitSingh11, line 131: I think we should not remove the OR conditon because there may be some customer who may be in older version of wget, this is a reason we added the OR condition. Also number of retries was 10 from a very long time, is it okay to reduce it to 3 ?

ShobhitSingh11: Ok for number of retires it makes sense as we have fallbacks, but we may need to see the OR part being removed

Sapa96: added 4 retries instead and added similar condition in the fallback to support older wget versions

ShobhitSingh11 (approved):
LGTM, get review from @devkimittal or @raghavharness for number of retries

raghavharness (approved):

PR #644, opened 2025-08-05
feat: [CI-18088]: enable authenticated Git operations (like push or f…

ebtasam-faridy:
JIRA: https://harness.atlassian.net/browse/CI-18088
We want to enable authenticated Git operations (like push or fetch) in later Run steps of a pipeline.
The drone-git plugin already stores Git credentials in a .netrc file during the initial clone step.
Our goal is to share this .netrc file with all containers in the Pod.
To do this, we create an emptyDir volume (netrc-volume) that acts as shared storage.
An init container creates the .netrc file, which drone-git populates.
Each container mounts this volume so .netrc appears at the right paths (/root/.netrc and /home/drone/.netrc).
This allows seamless, reusable Git authentication across all steps without re-authentication.

# Commit Checklist

## The Basics

### DIFF
--- a/app/oshelp/oshelp.go
+++ b/app/oshelp/oshelp.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/dchest/uniuri"
+	"github.com/harness/lite-engine/engine/spec"
 
 	"github.com/drone/runner-go/shell/bash"
 	"github.com/drone/runner-go/shell/powershell"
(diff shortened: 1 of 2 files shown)


app/oshelp/oshelp.go
```diff
@@ -66,6 +67,30 @@ func GetNetrc(os string) string {
...
 
+func GetNetrcFile(os string, env map[string]string) *spec.File {
+	netrcName := GetNetrc(os)
+
+	var path string
+	switch os {
+	case OSLinux:
+		path = fmt.Sprintf("/home/ubuntu/%s", netrcName)
+	case OSWindows:
+		path = fmt.Sprintf(`C:\Windows\system32\config\systemprofile\%s`, netrcName)
```
devkimittal, line 78: where did we get this path?

app/oshelp/oshelp.go
```diff
@@ -66,6 +67,30 @@ func GetNetrc(os string) string {
...
+	case OSMac:
+		path = fmt.Sprintf("/Users/anka/%s", netrcName)
```
devkimittal, line 80: are you sure this will be the path everytime? what if the customer is using vm runner that has some other provider than anka?

PR #664, opened 2025-08-26
feat: [CI-18759]: Improve logging and additionally added hotpool related logs

raghavharness:
1. Added logs in json format which will help in querying in cloud 
2. Used internal logger for any internal logs and kept UI logger separate.
3. Added hot pool logs to get information whether warm vm was picked or not.

### DIFF
--- a/app/drivers/manager.go
+++ b/app/drivers/manager.go
@@ -387,6 +387,13 @@ func (m *Manager) Provision(
 		return nil, fmt.Errorf("provision: failed to list instances of %q pool: %w", poolName, err)
 	}
 
+	logger.FromContext(ctx).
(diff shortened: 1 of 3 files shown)


abhay084 (approved):

app/drivers/manager.go
```diff
@@ -387,6 +387,13 @@ func (m *Manager) Provision(
 		return nil, fmt.Errorf("provision: failed to list instances of %q pool: %w", poolName, err)
 	}
 
+	logger.FromContext(ctx).
```
devkimittal, line 390: do we need log? it will cause too many logs

devkimittal: what's a pool

PR #677, opened 2025-09-16
feat: [CI-18946]: Enable global pool support

raghavharness:
This PR adds support for enabling global pool.
With this any pod can use a VM provisioned by any other pod.

### DIFF
--- a/app/drivers/distributed_manager.go
+++ b/app/drivers/distributed_manager.go
@@ -135,6 +135,10 @@ func (d *DistributedManager) cleanPool(ctx context.Context, pool *poolEntry, que
 	}
 
 	// 2. Destroy the claimed instances
+	logrus.WithField("pool", pool.Name).
+		WithField("instance_count", len(instancesToDestroy)).
+		WithField("instances", instancesToDestroy).
+		Infoln("cleaning up instances")
 	err := pool.Driver.Destroy(ctx, instancesToDestroy)
 	if err != nil {
 		// If destroy fails, we don't proceed to delete them from the DB.
--- a/app/drivers/distributed_manager.go
+++ b/app/drivers/distributed_manager.go
@@ -2,15 +2,19 @@ package drivers
 
 import (
 	"context"
+	"database/sql"
+	"errors"
 	"fmt"
 	"runtime/debug"
 	"time"
 
 	"github.com/Masterminds/squirrel"
 
+	"github.com/drone-runners/drone-runner-aws/command/harness/common"
 	"github.com/drone-runners/drone-runner-aws/store"
 	"github.com/drone-runners/drone-runner-aws/types"
 	"github.com/drone/runner-go/logger"
+	"github.com/harness/lite-engine/engine/spec"
 
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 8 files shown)


raghavharness:
Screenshot 2025-09-17 at 12 27 04 PM
Screenshot 2025-09-17 at 12 27 20 PM

app/drivers/distributed_manager.go
```diff
@@ -47,6 +94,78 @@ func (d *DistributedManager) CleanPools(ctx context.Context, destroyBusy, destro
...
+
+	// 2. Destroy the claimed instances
+	logrus.WithField("pool", pool.Name).
+		WithField("instance_count", len(instancesToDestroy)).
+		WithField("instances", instancesToDestroy).
+		Infoln("cleaning up instances")
+	err := pool.Driver.Destroy(ctx, instancesToDestroy)
+	if err != nil {
+		// If destroy fails, we don't proceed to delete them from the DB.
+		// The instances will remain in 'terminating' state for a later retry.
```
anurag-harness, line 145: How does this later retry happen?

app/drivers/distributed_manager.go
```diff
@@ -64,6 +183,192 @@ func (d *DistributedManager) IsDistributed() bool {
...
+	// Try to find and claim a free instance atomically
+	inst, err := d.instanceStore.FindAndClaim(ctx, &types.QueryParams{PoolName: poolName}, types.StateInUse, allowedStates)
+	if err != nil && err != sql.ErrNoRows {
+		return nil, false, fmt.Errorf("provision: failed to find and claim instance in %q pool: %w", poolName, err)
+	}
+
+	// If we successfully claimed an instance, update it and return
+	if inst != nil {
+		inst.OwnerID = ownerID
+		if inst.IsHibernated {
```
anurag-harness, line 211: Shouldnt we have types.StateHibernated in allowedStates above if we think we will hit this condition? We also use setupInstanceWithHibernate so we hibernate new instances set up in the pool right?

app/drivers/distributed_manager.go
```diff
@@ -190,10 +489,11 @@ func (d *DistributedManager) startInstancePurger(ctx context.Context, pool *pool
 
 	err = pool.Driver.Destroy(ctx, instances)
 	if err != nil {
-		return fmt.Errorf("distributed dlite: failed to delete instances of pool=%q error: %w", pool.Name, err)
+		logr.WithError(err).Errorf("distributed dlite: failed to delete instances of pool=%q", pool.Name)
```
HenAmar, line 492: can you explain why the return was removed from here?

app/drivers/distributed_manager.go
```diff
@@ -47,6 +94,78 @@ func (d *DistributedManager) CleanPools(ctx context.Context, destroyBusy, destro
...
+		allowedStates = append(allowedStates, types.StateCreated, types.StateHibernating)
+	}
+
+	// Set the pool name on the query parameters
+	query.PoolName = pool.Name
+
+	// 1. Find and claim all matching instances
+	var instancesToDestroy []*types.Instance
+	for {
+		// Claim one instance at a time to avoid long transactions and to process them sequentially.
```
HenAmar, line 117: we need a trade off here. For large pools, this creates many round-trips to the database and Increases contention when multiple runners are cleaning up simultaneously

app/drivers/distributed_manager.go
```diff
@@ -64,6 +183,192 @@ func (d *DistributedManager) IsDistributed() bool {
...
+				Warnln("hibernation not ready, retrying after delay")
+			time.Sleep(delay)
+			continue
+		}
+
+		// For max retries reached, break immediately
+		break
+	}
+
+	if hibernateErr != nil {
```
HenAmar, line 352: don't we have a race here? between the failure and the revert to StateCreated?

PR #680, opened 2025-09-20
fix: [CI-19086]: Change logic for BYOI to use local VMs instead of directly from OCI

raghavharness:
1. Now using a local VM for BYOI instead of directly cloning from OCI Image.
2. Removed logic to tweak keychain for login, directly using tart env vars now

### DIFF
--- a/app/drivers/nomad/mac_virtualizer.go
+++ b/app/drivers/nomad/mac_virtualizer.go
@@ -52,7 +52,7 @@ func (mv *MacVirtualizer) GetInitJob(vm, nodeID, userData, machinePassword, defa
 		return nil, "", "", err
 	}
 	encodedUserData := base64.StdEncoding.EncodeToString([]byte(uData))
-	startupScript := base64.StdEncoding.EncodeToString([]byte(mv.generateStartupScript(vm, machinePassword, defaultVMImage, vmImageConfig, resource, port)))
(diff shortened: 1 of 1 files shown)


app/drivers/nomad/mac_virtualizer.go
```diff
@@ -197,19 +208,28 @@ else
...
-	  security unlock-keychain -p "$MACHINE_PASSWORD" ~/Library/Keychains/login.keychain-db
-      echo "$REGISTRY_PASSWORD" | /opt/homebrew/bin/tart login "$REGISTRY" --username "$REGISTRY_USERNAME" --password-stdin
+	  echo "Pulling image $VM_IMAGE..."
+	  TART_REGISTRY_HOSTNAME="$REGISTRY" TART_REGISTRY_USERNAME="$REGISTRY_USERNAME" TART_REGISTRY_PASSWORD="$REGISTRY_PASSWORD" /opt/homebrew/bin/tart pull "$VM_IMAGE"
+
+      echo "Exporting image $VM_IMAGE to /tmp/${ENCODED_IMAGE_NAME}.tvm..."
+      /opt/homebrew/bin/tart export "$VM_IMAGE" "/tmp/${ENCODED_IMAGE_NAME}.tvm"
+
+	  echo "Pruning tart OCI cache..."
+	  /opt/homebrew/bin/tart prune --entries caches --older-than=0
```
rajatharanganath, line 218: Review in follow-up PR for handling race conditions.

PR #842, opened 2026-04-17
fix: [CI-22126]: findInstanceZone searches all configured zones (pool + networkConfigs)

raghavharness:
## Problem
`findInstanceZone` in `app/drivers/google/driver.go` iterates `p.zones` directly when looking up which zone an instance lives in. When the GCP driver is configured with `networkConfigs[]` (multi-network setup), networks can declare additional zones that are not present in `p.zones`. Instances created in those networkConfig-only zones are reported as `ErrInstanceNotFound` even though they exist, breaking `Logs`, `Destroy`, hibernation, and orphan cleanup paths.

## Root Cause
```go
for _, zone := range p.zones { ... }
```
ignores `p.networkConfigs[].zones`. The helper `allZones()` (driver.go:186) already exists for exactly this purpose (pool zones + networkConfig zones, deduplicated).

## Solution
Replace the iteration with `p.allZones()` so `findInstanceZone` considers every zone the driver is configured to place instances in.

## Changes Made
- `app/drivers/google/driver.go`: `findInstanceZone` now iterates `p.allZones()` instead of `p.zones`.

## Testing
- `go build ./...` ✅
- `go test ./app/drivers/google/... -count=1` ✅
- `go vet ./app/drivers/google/...` ✅

No behavior change when `networkConfigs` is empty — `allZones()` falls back to `p.zones` in that case (driver.go:187-189).

## Related Issues/Logs
- JIRA: [CI-22126](https://harness.atlassian.net/browse/CI-22126)

---

[CI-22126]: https://harness.atlassian.net/browse/CI-22126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/drivers/google/driver.go
+++ b/app/drivers/google/driver.go
@@ -1133,7 +1133,7 @@ func (p *config) findInstanceZone(ctx context.Context, instanceID string) (
 	var lastErr error
 	allNotFound := true
 
-	for _, zone := range p.zones {
+	for _, zone := range p.allZones() {
 		_, err := p.getInstance(ctx, p.projectID, zone, instanceID)
 		if err == nil {
 			return zone, nil


raghavharness:
trigger messagecheck

tapankarangiya (approved):

PR #853, opened 2026-04-29
feat: [CI-22397]: store firewall project in DB for correct purger deletion

tapankarangiya:
## Problem
The firewall purger (`PurgeOrphanedFirewallRules`) passes `nil` for instance to `CleanupEgressPolicy`, so the Google driver always falls back to the default pool project (`p.projectID`). In Shared VPC setups where firewalls are created in the host project, this causes deletion to fail.

##Testing 
Firewalls are being created and deleted and project is getting stored in firewall_rules table
Screenshot 2026-04-29 at 3 34 12 PM

## Root Cause
`CleanupEgressPolicy` accepted `*types.Instance` but only the Google driver used it — and only to extract a project string from `instance.Network`. All 8 other drivers ignored the parameter entirely. The purger has no instance available, so it passed `nil`.

## Solution
Replaced `*types.Instance` with `project string` in the `Driver.CleanupEgressPolicy` interface. The firewall project is now:
- Resolved from `instance.Network` at rule creation time and stored in the `firewall_rules` table
- Passed directly by the purger from the DB record
- Resolved from `instance.Network` at per-instance destroy time

## Changes Made
- **`types/types.go`**: Added `Project` field to `FirewallRule` struct; moved `ProjectFromNetwork` here to avoid circular imports (`google` → `drivers` → `google`)
- **`app/drivers/pool.go`**: Changed `CleanupEgressPolicy` signature: `*types.Instance` → `project string`
- **`app/drivers/google/driver.go`**: Updated implementation to use `project` param directly; `ProjectFromNetwork` now delegates to `types.ProjectFromNetwork`
- **`app/drivers/instance_ops.go`**: Purger passes `rules[0].Project`; per-instance destroy resolves project via `types.ProjectFromNetwork`
- **`command/harness/setup.go`**: Stores resolved firewall project when creating rules
- **`store/database/sql/firewall_rules.go`**: Added `project` to insert/select queries
- **DB migrations**: Added `project` column to `firewall_rules` table (postgres + sqlite)
- **8 no-op drivers + 3 test mocks**: Updated signature

## Testing
- `go build ./...` ✅
- `go test ./... -count=1` ✅
- `go vet ./...` ✅

## Related Issues/Logs
- Follows up on [CI-22369](https://harness.atlassian.net/browse/CI-22369) which fixed cross-project firewalls for the per-instance destroy path but not the purger path

---

[CI-22369]: https://harness.atlassian.net/browse/CI-22369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tapankarangiya:
trigger messagecheck

store/database/migrate/sqlite/0015_add_project_to_firewall_rules.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE firewall_rules ADD COLUMN project VARCHAR(250) NOT NULL DEFAULT '';
```
raghavharness: use project_id everywhere for naming

tapankarangiya: done

command/harness/setup.go
```diff
@@ -620,19 +620,23 @@ func applyAndSaveEgressRules(
 	allowRuleName := google.EgressRuleName(google.EgressAllowPrefix, instance.ID)
 	denyRuleName := google.EgressRuleName(google.EgressDenyPrefix, instance.ID)
 
+	// Resolve the firewall project from the instance's network so the purger can delete rules
+	// from the correct project later (Shared VPC firewalls live in the host project, not the VM project).
+	firewallProject := google.ProjectFromNetwork(instance.Network, "")
+
```
raghavharness, line 626: maybe for later but should make this generic and not google.

raghavharness (approved):

PR #857, opened 2026-05-04
feat: [CI-22397]: Remove SkipCloudInitPackages dependency; rely on IsHosted to skip cloud-init package install

saurabh-harness:
# Description
Remove `SkipCloudInitPackages` dependency; rely on `IsHosted` to skip cloud-init package install

## Details
`SkipCloudInitPackages` is a runner-level toggle (env var `DRONE_SKIP_CLOUD_INIT_PACKAGES`, wired through `manager_config.go `-> `manager.go` -> `provisioner.go` -> `lehelper` -> `cloudinit.go`) that controls whether the generated cloud-init script installs packages (docker, git, wget, etc.) during VM bootstrap. It was originally introduced as a gradual rollout flag so hosted envs could opt out of package installation (images are pre-baked with all required binaries, so installing them again is pure boot-time overhead).

This flag is now enabled in all environments (prod, qa, prod-free, etc.). The `IsHosted` flag already travels on createOptions alongside it, and the two are effectively coupled — any hosted flow wants packages skipped, any non-hosted flow does not.

# Goal
Remove `SkipCloudInitPackages` from the codebase and make the cloud-init package-install logic gate on `IsHosted` directly.

# Tested via
#### No occurrence of `SkipCloudInitPackages`
Screenshot 2026-05-04 at 3 34 26 PM

#### Build passed
Screenshot 2026-05-04 at 3 35 20 PM


app/cloudinit/user_data/ubuntu_linux
```diff
@@ -1,7 +1,7 @@
 #cloud-config
-{{ if not .SkipCloudInitPackages }}
+{{ if not .IsHosted }}
 # Package installation
-{{ if and (.IsHosted) (eq .Platform.Arch "amd64") }}
+{{ if (eq .Platform.Arch "amd64") }}
 packages:
   - wget
```
raghavharness: we should remove the if cond and only keep else part

raghavharness (approved):

PR #860, opened 2026-05-05
feat: [CI-22316]: Add instance source (pool/predictor/ondemand) as a VM label for hosted instances

saurabh-harness:
# Description
Add instance source (pool/predictor/ondemand) as a VM label for hosted instances

# Goal
Include the instance source as a VM label alongside pool_id and harness_env when the instance is hosted, so the label is visible directly on the GCP instance (UI, billing exports, log labels, etc.) without a DB lookup.

# Tested via
- [x] `source` is added in the **labels**
- [x] `harness_env` and `pool_id` were consistent

## Before
image

## After
image

### DIFF
--- a/app/drivers/provisioner.go
+++ b/app/drivers/provisioner.go
@@ -268,9 +268,11 @@ func (m *Manager) setupInstance(
 		retain = "true"
 	}
 	createOptions.InternalLabels = map[string]string{"retain": retain}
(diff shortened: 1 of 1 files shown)


saurabh-harness:
trigger messagecheck

raghavharness (approved):

PR #865, opened 2026-05-27
fix: [CI-22929] log LE HTTP request payload on transport failure

avijeetsm:
## What

Adds debug instrumentation to surface the outgoing HTTP request whenever a
call from the runner to lite-engine (LE) fails at the transport layer
(connection refused, TLS handshake error, timeout, EOF, etc.).

Also bumps `github.com/harness/lite-engine` from `v0.5.171` → `v0.5.175`,
along with the matching default download URLs in `DRONE_LITE_ENGINE_PATH`
and `DRONE_LITE_ENGINE_FALLBACK_PATH` so VMs install the LE version the
runner is built against.

## Why

CI-22929: LE HTTP calls (health/setup/poll-step/start-step) intermittently
fail and today the logs only show the error string — there's no visibility
into the request that failed (URL, headers, body), which makes
transport-level issues hard to triage.

## How

Wraps the `*http.Client.Transport` returned by `lehttp.NewHTTPClient` with
a `loggingTransport` (an `http.RoundTripper` decorator) in
`app/lehelper/lehelper.go`. On a failed `RoundTrip`, it dumps the request
via `httputil.DumpRequestOut` and logs it alongside the error at `Debug`.
Because every LE call goes through `http.Client.Do → Transport.RoundTrip`,
every existing call site picks this up automatically — no API changes.

## Scope

- `app/lehelper/lehelper.go` — add `loggingTransport`, wire into `GetClient`
- `command/config/config.go` — bump default LE download URLs to v0.5.175
- `go.mod` / `go.sum` — bump `github.com/harness/lite-engine` to v0.5.175

## Risk / Rollout

- Logging only fires on transport errors, so happy-path overhead is one
  extra `RoundTrip` indirection per LE call.
- No behavioral change on the request itself; the wrapped transport's
  response and error are returned unchanged.
- Intended as a debug aid for CI-22929; can be reverted independently of
  the LE bump.

## Test plan

- [ ] Build runner with this change and run an existing CI job — confirm
      no regression on a successful LE setup/poll-step flow.
- [ ] Force a failure (e.g. block port 9079 on the VM, or use a bad cert)
      and confirm the request dump + error appear in the runner logs at
      Debug level.
- [ ] Verify `v0.5.175` LE binary is downloaded by VMs using default config.

### DIFF
--- a/app/lehelper/lehelper.go
+++ b/app/lehelper/lehelper.go
@@ -2,15 +2,37 @@ package lehelper
 
 import (
 	"fmt"
+	"net/http"
+	"net/http/httputil"
 	"time"
 
 	"github.com/harness/lite-engine/api"
 	lehttp "github.com/harness/lite-engine/cli/client"
+	"github.com/sirupsen/logrus"
 
 	"github.com/drone-runners/drone-runner-aws/app/cloudinit"
 	"github.com/drone-runners/drone-runner-aws/app/oshelp"
 	"github.com/drone-runners/drone-runner-aws/types"
 )
+
+type loggingTransport struct {
+	wrapped http.RoundTripper
+}
+
+func (t *loggingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+
+	resp, err := t.wrapped.RoundTrip(req)
+	if err != nil {
+		dump, _ := httputil.DumpRequestOut(req, true)
+		logrus.WithField("request", string(dump)).Traceln("LE HTTP request")
+		logrus.WithError(err).Errorln("LE HTTP request failed")
+		return resp, err
+	}
+
+	dumpResp, _ := httputil.DumpResponse(resp, true)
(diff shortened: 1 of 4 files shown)


app/lehelper/lehelper.go
```diff
@@ -2,16 +2,38 @@ package lehelper
...
+	resp, err := t.wrapped.RoundTrip(req)
+	if err != nil {
+		dump, _ := httputil.DumpRequestOut(req, true)
+		logrus.WithField("request", string(dump)).Traceln("LE HTTP request")
+		logrus.WithError(err).Errorln("LE HTTP request failed")
+		return resp, err
+	}
+
+	dumpResp, _ := httputil.DumpResponse(resp, true)
+	logrus.WithField("response", string(dumpResp)).Traceln("LE HTTP response")
```
raghavharness: we need this ?

PR #874, opened 2026-06-19
feat:[CI-23143]:enable retry of vm creation if vm creation fails

avijeetsm:
# Commit Checklist

## The Basics

### DIFF
--- a/app/drivers/google/driver.go
+++ b/app/drivers/google/driver.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/dchest/uniuri"
 	"github.com/google/uuid"
+	"github.com/hashicorp/golang-lru/v2/expirable"
 	"google.golang.org/api/compute/v1"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/option"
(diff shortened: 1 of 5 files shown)


raghavharness (commented):
## Summary

This PR adds zone/network stockout-retry to the GCP driver's `create` path (`app/drivers/google/driver.go`). It builds an ordered list of `createCandidate`s (network configs in declared order, zones shuffled within each), then `insertWithStockoutRetry` walks them, retrying alternate candidates only on stockout errors (`isStockoutError`) and remembering exhausted `(zone, machineType)` pairs in an expirable LRU so future creates deprioritize them. This addresses [CI-23143](https://harness.atlassian.net/browse/CI-23143) where round-robin fallback re-picked a stocked-out zone. Retry is disabled when a persistent disk or capacity reservation is used.

## Verdict

**LGTM with nits** — see below. Well-scoped change with thorough unit coverage for candidate ordering and stockout detection.

## Concerns

- **Question:** On success, only `zone` and `resolvedNetwork` are reassigned from `succeeded` ([driver.go \~688](https://github.com/drone-runners/drone-runner-aws/pull/874/files)); `resolvedSubnetwork`/`resolvedTags` aren't. If anything downstream in `create` references those after a retry to a different candidate, they'd be stale — worth confirming they're unused past this point.
- **Suggestion:** `maxStockoutAttempts = 3` caps total attempts (first + 2 alternates), so deep multi-network/zone pools only ever try 3. Confirm that's the intended retry breadth vs. pool fallback in `HandleSetup`.
- **Nit:** `insertWithStockoutRetry` mutates the shared `in` spec per attempt; tags are correctly copied, but a brief comment noting the spec is intentionally reused (not rebuilt) per attempt would help future readers.

avijeetsm:
@raghavharness Only zone/network survive into the returned instance; subnetwork & tags were already applied to the spec per-candidate inside insertWithStockoutRetry.
3 retry attemps is intentional. I do not feel that it would be possible, except under very rare circumstances, to have same machine stock out in 3 different zones. At that point, it would be prudent to check some other issue than to attempt on 4th zone.

raghavharness (approved):

PR #890, opened 2026-07-14
feat: [CI-23727]: remove legacy TPA-based egress implementation in favor of HTTP proxy

raghavharness:
## Problem

The drone-runner-aws codebase maintained two parallel egress control implementations:

1. **v1 (TPA-based):** Used `TPAAddress`, `TPAPort`, and `EgressProxyEnabled` flag to configure Envoy proxy with MITM interception via the `hosted_ubuntu_linux_egress` template
2. **v2 (HTTP proxy):** Uses standard `EgressProxyURL`, `EgressNoProxy`, `EgressCACert` with the `ubuntu_linux_egress_v2` template

The v1 implementation added complexity and required maintaining separate cloud-init templates, conditional logic, and additional database fields. The v2 approach is simpler, more standard, and actively maintained.

## Root Cause

Legacy code from the initial egress control implementation that predated the simplified HTTP proxy approach. The v1 TPA-based implementation was superseded by v2 but never removed.

## Solution

Remove the legacy v1 TPA-based egress implementation completely in favor of the v2 HTTP proxy approach.

## Changes Made

**Core Changes:**
- Deleted `app/cloudinit/user_data/hosted_ubuntu_linux_egress` template (136 lines)
- Removed `TPAAddress`, `TPAPort`, `EgressProxyEnabled` fields from `cloudinit.Params` struct in `types/types.go`
- Simplified `cloudinit.go` logic to unconditionally use `ubuntuEgressProxyTemplate` when `EgressControl=true`
- Removed conditional branches throughout the codebase that checked `EgressProxyEnabled`

**Test Updates:**
- Updated `TestLinuxEgressDispatch` to verify v2 template selection exclusively
- Removed `TestLinuxEgressProxyDispatch` tests for v1 vs v2 comparison (no longer needed)
- Updated `TestUbuntuBinariesPartialShared` and `TestWindowsBinariesPartialShared` to remove v1 test cases
- All tests now verify the absence of v1 Envoy markers (`systemctl enable --now envoy-proxy`)

**Database Migrations:**
- Added `0025_update_table_instances_proxy_url.up.sql` (PostgreSQL) to consolidate proxy URL handling
- Added `0016_update_table_instances_proxy_url.up.sql` (SQLite) for the same purpose

**Files Changed:** 21 files, 203 insertions(+), 351 deletions(-)

## Testing

✅ All existing tests pass (`go test ./... -count=1`)
✅ Build succeeds (`go build ./...`)
✅ Pre-push hook validated all tests before push
✅ Cloud-init tests verify v2 template is used exclusively when `EgressControl=true`
✅ Tests confirm v1 Envoy markers are absent from rendered output

## Related Issues/Logs

**JIRA:** https://harness.atlassian.net/browse/CI-23727

---


raghavharness:
trigger messagecheck

tapankarangiya (approved):
LGTM — clean removal of the v1 TPA/envoy-MITM flow. Build is clean, tests pass, migrations and column ordering in `store/database/sql/instances.go` line up correctly with the new `instance_proxy_url` column.

A few non-blocking things worth flagging:

1. **Behavioural change for existing deployments**: previously enablement required `DRONE_EGRESS_PROXY_ENABLED=true` AND `egress_control=true` on the pool. Now pool `egress_control=true` alone triggers the v2 proxy template + step env injection. Any deployment that had `egress_control: true` on pools but was intentionally running with `DRONE_EGRESS_PROXY_ENABLED=false` will now start injecting proxy env vars. Assuming this is intentional (it's the whole point of the deprecation), it should be called out in the release notes.

2. **`GoogleNetwork.Tags` deprecation asymmetry** (`command/config/config.go:229`): top-level `Tags` is commented as deprecated but the poolfile still unconditionally calls `google.WithTags(g.Tags...)` — no warning like the one added for `Network`/`Subnetwork` at `app/poolfile/config.go:223`. Minor consistency nit.

3. **Migration is additive-only, no down**: matches existing pattern in the repo (no `down.sql` files present), so fine, but worth noting rollback would need a manual `ALTER TABLE DROP COLUMN`.

4. **`optionalWhenCheckAll` map** (`command/harness/common/common.go:52`) is allocated per call to `ValidateStructForKeys`. Trivial, but easily hoisted to a package-level var if this ever gets called in a hot path.

PR #891, opened 2026-07-15
feat:[CI-23780]:different timeout for free pool

avijeetsm:
# Commit Checklist

## The Basics

### DIFF
--- a/app/drivers/google/driver.go
+++ b/app/drivers/google/driver.go
@@ -387,9 +387,13 @@ func (p *config) ReserveCapacity(ctx context.Context, opts *types.InstanceCreate
 		Description:                 fmt.Sprintf("Capacity reservation for pool %s", opts.PoolName),
 	}
 
-	if opts.CapacityReservationTTL > 0 {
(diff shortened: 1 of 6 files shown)


anurag-harness (changes requested):
This wont help much. GCP does not give a guarantee that it will delete the reservation right after the timeout. It takes upto 2 hours to delete it. We can think of other ways to handle this like
1. Making the purger run more frequently (every 10 mins)
2. Reducing the time we consider a capacity as stale from 20 mins to 10 mins
3. Modifying the cloud run job we run to delete capacity to run more frequently and to remove any unused capacity over 10 mins
4. Increasing the timeout given to reserve capacity from 5 mins to 10 mins. This will reduce the chance of orphaned reservations when we try fallback pools (we can discuss this)

anurag-salesforce:
There is still no point of this PR. It just sets the timeout in GCP which is already reduced to 10 mins: https://github.com/drone-runners/drone-runner-aws/commit/ab47b601a3861ccbbfa31d23b028ac9f8e9b6bc7

PR #898, opened 2026-07-23
feat:[CI-23929]:add metrics for gcp APIs

avijeetsm:
# Commit Checklist

## The Basics

### DIFF
--- a/app/drivers/google/driver.go
+++ b/app/drivers/google/driver.go
@@ -21,6 +21,7 @@ import (
 	"github.com/drone-runners/drone-runner-aws/app/oshelp"
 	itypes "github.com/drone-runners/drone-runner-aws/app/types"
 	"github.com/drone-runners/drone-runner-aws/command/harness/storage"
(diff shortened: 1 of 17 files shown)


raghavharness:
IMO, we should not call the metrics to be GCP specific. It should be cloud agnostic. cloud_provider should be like a filter which can be part of these metrics. Let me know your thoughts.

chhawchharia:
> IMO, we should not call the metrics to be GCP specific. It should be cloud agnostic. cloud_provider should be like a filter which can be part of these metrics. Let me know your thoughts.

+1

avijeetsm:
> IMO, we should not call the metrics to be GCP specific. It should be cloud agnostic. cloud_provider should be like a filter which can be part of these metrics. Let me know your thoughts.


@raghavharness 
There are more non gcp specific metrics too. Some of these have to be GCP specific. for example things like API call rate, stockouts etc. these things are vendor specific. these vendor specific metrics are only in the gcp driver

